PR #49, opened 2015-01-08
Augmenting the Permission Model documents

jeremyf:
[skip ci]

### DIFF
--- a/artifacts/permission_model.md
+++ b/artifacts/permission_model.md
@@ -1,28 +1,48 @@
-# Permission Model (Draft v1)
+# Permission Model (Draft v2)
 
 **What follows is provisional, full of likely inaccuracies, but is meant as a guide post for understanding the authorization model.**
 
 ## Glossary
 
 * Actor: A Group or User.
+* Action: A thing that will/can be done to an object.
+* AuthorizationLayer: The layer that asserts can a User perform the Action on the Entity.
 * Entity: A permissionable entity.
+* EntityType: The Rails base_class of an object; Needed for polymorphic relationships
 * Group: A name for a group of users. Think in terms of nouns.
 * Group Membership: Association of a user to a group.
 * Permission: Intersection of an Entity, Actor, and Role
 * Policy: Answers the question: Can the User take the given action on the given Entity.
-* Role: A name for a set of responsibilities. Think in terms of verbs.
+* PolicyQuestion: A misnomer, perhaps better reflected as an action. See Action.
+* Role: A name for a set of actions that can performed. Think in terms of verbs.
+* Runner: Responsible for coordinating a User performing one or more Action/Entity pairs
 * User: A person.
+* WorkType: examples include ETD, SeniorThesis, Image; This is not the same as EntityType
 
-### Example:
+## Concerns
 
-Assume the following:
+There are two primary concerns in the Authorization subsystem.
 
-* Alice is a User
-* Cataloging is a Role
-* Catalogers is a Group
-* Alice is a member of the Catalogers group
-* BigBadBook is an Entity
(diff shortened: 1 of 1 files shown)


dbrower:
Some places role is used where it seems "action" is intended.

I'm curious how role really fits in here.

I have an (possibly bad) idea. The atomic elements are Actor(U), Action(A), Item (I). The global permissions are a subset of the triples U * A * I. Writing out each triple is tedious. Groups let us write a bunch of triples like: given action a and item i, we get (G,a,i) = (u,a,i) for all u in G.

It seems a role should be a subset of Actions*Items so that giving a user u the role R we are specifying  the triples (u,R) = (u,a,i) for all (a,i) in R.

I don't know how that would work in practice, but it seems to make sense conceptually.

jeremyf:
This is the loose model I was lifting:

https://github.com/stffn/declarative_authorization#authorization-data-model

jeremyf:
@dbrower  I believe your observations are correct. I'm wondering, however, if there should be something along the lines of:

```
(u,X,i) = (u,a,i) for all a in X.
```

jeremyf:
However, the documentation at this point is descriptive and not prescriptive.

jeremyf:
@dbrower One thing to consider is that actions on an entity are available by state as well. The following hash is in effect as users are working on a given ETD.

```ruby
      STATE_POLICY_QUESTION_ROLE_MAP =
      {
        'new' => {
          update?: ['creating_user', 'advisor'], show?: ['creating_user', 'advisor'],
          delete?: ['creating_user'], submit_for_review?: ['creating_user']
        },
        'under_review' => {
          update?: ['etd_reviewer'], show?: ['creating_user', 'advisor', 'etd_reviewer'],
          request_revisions?: ['etd_reviewer'], approve_for_ingest?: ['etd_reviewer']
        },
        'ready_for_ingest' => { show?: ['creating_user', 'advisor', 'etd_reviewer'] },
        'ingested' => { show?: ['creating_user', 'advisor', 'etd_reviewer'] },
        'ready_for_cataloging' => { show?: ['creating_user', 'advisor', 'etd_reviewer', 'cataloger'], finish_cataloging?: ['cataloger'] },
        'cataloged' => { show?: ['creating_user', 'advisor', 'etd_reviewer', 'cataloger'] },
        'done' => { show?: ['creating_user', 'advisor', 'etd_reviewer', 'cataloger'] }
      }.freeze
```

https://github.com/ndlib/sipity/blob/master/app/state_machines/sipity/state_machines/etd_student_submission.rb#L15-L31

dbrower:
@jeremyf thanks for the link to the gem. It is implementing Role Based Access Control, which is a different thing than what is described here...as much as I can tell. The existence of these models is new to me. In RBAC each user is given a set of roles (which would be analogous to actors in this description). Each role describes what actions can be done and which objects those actions can be done on. In the defining paper (Ferraiolo and Kuhn, "Role-Based Access Controls", 1992) there is a variant which makes the object explicit.

But the data and operations can be subject to extra conditions, so it is possible to only allow, say, an author to edit his own works.

dbrower:
I think the (u,X,i) idea is a little weak, because of where will this be stored? with a user? with an item? what does it even mean? it grants a bunch of actions to a specific user-item pair.

Tying actions to a state machine can be done...i think invalid actions should be done at a different layer than the authorization model. it seems to be the state machine's responsibility to ignore irrelevant actions.

dbrower:
I don't see why this PR should  be held up, but this is a good conversation place. How would you like to proceed?

jeremyf:
@dbrower Merge it. I'm reworking a few things to help create the necessary seem for improvements.

PR #51, opened 2015-01-08
Persisting Sip#work_type

jeremyf:
Following on sipity@e0e1a8a, I want to make sure that we are capturing
work_type and persisting it. The current implementation is naive in
that it assumes only one work type. However, we will certainly be
adding additional ones over the coming weeks.


rbalekai:
:+1:

PR #52, opened 2015-01-08
Extracting mapping function for group name to role

banurekha:


jeremyf:
I'm going to resubmit a different pull request. There are enough moving parts. And I want to clean some things out.

PR #55, opened 2015-01-08
Spec Query#group_names_for_entity_and_acting_as

jeremyf:
There is a naive hash that translates a role name to a group name.
I want to expose a query method to retrieve that information; This
information will most certainly make it into a database for greater
configuration options.

Also, consider the fact that the Sipity::Commands::PermissionCommands
module will need to be changed and the underlying spec would benefit
from a tweak as well, saying that the Sipity::Queries::GroupQueries
is allowed to receive #group_names_for_entity_and_roles and return a
specified value.

See Sipity::Commands::PermissionCommands#grant_groups_permission_to_entity_for_role

[skip ci]

Refines to #52

### DIFF
--- a/app/repositories/sipity/commands/permission_commands.rb
+++ b/app/repositories/sipity/commands/permission_commands.rb
@@ -18,11 +18,8 @@ module Sipity
       #
       # @raise Exception if for any of the given acting_as, no group could be found
       def grant_groups_permission_to_entity_for_acting_as!(entity:, acting_as:)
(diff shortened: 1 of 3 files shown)


banurekha:
I already worked renaming role to acting_as and will submit pull request shortly and I think this pull request not be necessary

jeremyf:
Looking to resolve #57 and this one.

jeremyf:
A duplicate of #57

PR #71, opened 2015-01-13
Create AccessRight model

jeremyf:
I opted to create a separate model as both a Sip and an
associated Attachment could have different access rights. I also went
the route of implementing a date range to reflect that as we deal with
Embargo's and Lease's the data gets convoluted as we attempt to
reflect two concepts (rights and dates) in one user input.


banurekha:
Do you think schema diagram change will be separate request

jeremyf:
Good catch. I'll add another commit to include the schema change.

I'm also going to pursue an ERD auto-generation option because this is something I'll keep forgetting.

PR #72, opened 2015-01-13
Adding Transient Answer model

jeremyf:
There are some answers that we ask our patrons to complete that are
prompts for additional thought or consideration. This model is the
persistence of those answers, and is to be used to continue prompting
for more permanent information.

Consider the question about Access Rights. There are three somewhat
simple answers: Open, Restricted, and Private. However with the
concept of Embargos and Leases, the actual answer can require a whole
lot more prompting; namely when does one access right start and the
other end?

### DIFF
--- a/app/models/sipity/models/sip.rb
+++ b/app/models/sipity/models/sip.rb
@@ -14,6 +14,7 @@ module Sipity
       has_many :attachments, foreign_key: :sip_id, dependent: :destroy
       has_one :doi_creation_request, foreign_key: :sip_id, dependent: :destroy
 
+      # TODO: Extract to TransientAnswer
       ALREADY_PUBLISHED = 'already_published'.freeze
       WILL_NOT_PUBLISH = 'will_not_publish'.freeze
       GOING_TO_PUBLISH = 'going_to_publish'.freeze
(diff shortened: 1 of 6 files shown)


banurekha:
how is this different from access_rights? could be list the option from access_right and store the answer sip directly?

jeremyf:
I amended the commit message with the following, @banurekha please review to see if it answers your questsion:

Adding Transient Answer model

There are some answers that we ask our patrons to complete that are
prompts for additional thought or consideration. This model is the
persistence of those answers, and is to be used to continue prompting
for more permanent information.

On the initial page for creating a sip we want to ask the following:
What is the applicable Access Right?

Possible Answers:
  * Open Access
  * Restricted
  * Private
  * It will change over time

In the case of Open Access, Restricted, and Private the answer is very
simple; We make an entry in AccessRight that starts as of today for
the given answer.

However, if the answer is "It will change over time", we don't want
to present additional decision points immediately, but instead want
to persist an answer that will require additional follow-up as we
assist our patron in completing their SIP.

The follow-up to the answer would allow for us to introduce a complex
decision in a more user focused and "gentle" manner. The idea being
that the process of depositing something should be a conversation.

PR #83, opened 2015-01-14
Reworking module_function declaration

jeremyf:
I am running into a case where the Ruby beautifier (and to some extent
style guide) are not liking the module_function with method def. What
happens is the auto-beautify is getting very confused. I use the auto-
beautify to make sure spacing is "correct" and there is no trailing
white-space.


rbalekai:
:+1:  waiting for travis to pass

PR #88, opened 2015-01-16
Extracting form commands to relevant form objects

jeremyf:
By extracting the submit behavior to the form, a few things will be
possible.

* Polymorphism related to forms - That is to say each form has a
  reasonable idea of what it needs to do; so do that. This aligns
  better with the Reform and ActiveForm behaviors
* Begin removing repository command module_functions - Given that many
  of the command behaviors were both a module_function and an instance
  method, it implies a certain complexity and leakyness in the
  abstraction.
* Removes several methods from the command modules
* Pushes the relevant form code closer to the object in question.

> Banu: so you are moving away from commands
>   module?
> Jeremy: In the case of the form's submit method, yes;
>   However in the case of interacting with the data, no. So there
>   will still be commands. Forms will compose the various required
>   commands. I can amend the commit if need be.
> Banu: you mean all form action belong to form
>   module rather than separate module.
> Jeremy: The form's submit method will call the repository
>   commands to persist the things that need to be persisted…or launch
>   the jobs that need to be launched.
> Banu: so how do you envision non-web interaction? they still need to
>   create a form object and submit
> Jeremy: Yup. The concept of the form is there for validation and
>   exposing only the required attributes. This is the first one.
>   There are a few others that I'll step through as well.
> Banu: make sense.. you removed citation commands because it was
>   doing only that


rbalekai:
:+1:

rbalekai:
waiting for tests to pass

PR #119, opened 2015-02-04
First pass at styling action items

danhorst:
@jeremyf it would be nice to have a #done? to style with on the todo items.


jeremyf:
Noted. And something we can address after this change is pushed up.

PR #126, opened 2015-02-05
Splicing in permission granting

jeremyf:
## Splicing in and deprecating permission granting

@b65264358d1faee7885d31b0a27dbcef78b81113

I want to leverage both systems at the same time, so I'm introducing
the appropriate seem.

## Adding Work#to_processing_entity

@5f2d9e4e53c1776da6d207da63cc64c3c7221544

This is something that I most certainly am going to use. So I'm laying
that out now.

Next up, convert_to_processing_strategy!

## Adding scenario regarding Models::Work

@17b0dc7e7dd84ff91350e31bc2c352ca11c94c90

This is a matter of confidence, and draws attention to its behavior.

## Adding conversion method for processing actor

@96891a5d5d907554c29feba58134c2317ff9fe90

I want to be able to ensure that when I'm working with a given user
or group that I'm transforming that into the other sub-system object.

## Adding method to convert string to a Role object

@172fc2da1430f70c228f09ca7e7b22b2498218b5

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -32,6 +32,7 @@ MethodLength:
   CountComments: false
   Enabled: true
   Exclude:
+    - app/conversions/**/*.rb
     - 'app/controllers/sipity/controllers/dois_controller.rb'
     - 'app/repositories/sipity/queries/permission_queries.rb'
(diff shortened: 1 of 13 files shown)


jeremyf:
Merged manually.

PR #128, opened 2015-02-06
Dltp 361 ToDo - Attach form

banurekha:
Can do the following

   Able to remove one or more of the files
    Add more additional files
    Mark one of the files as representative
    Added attribute/column on the Models::Attachment


jeremyf:
Now that there is a pid minter, use the files pid as the identifier instead of the file_name; Because file_name is not guaranteed to be unique.

jeremyf:
Also, the AttachForm has numerous methods that are untested. Add tests to demonstrate the expected interaction; All of those methods can be tested in in the attach_form_spec.

Also add for a test that is not UI based for submitting the form.

https://coveralls.io/builds/1880163/source?filename=app%2Fforms%2Fsipity%2Fforms%2Fwork_enrichments%2Fattach_form.rb

banurekha:
I think :remove_files and :mark_as_representative can be removed in place of :attachment_attributes. Although I am not sure how to validate only one of the files is representative and how to delete the files which are marked to delete

jeremyf:
Uniqueness validation is a bit bit tricky, however there are two things we can use to get the best approximation (without crafting another database table):

* Use a radio button for which one is representative; All of those buttons will need to have the same input name attribute in HTML (i.e. <input name="mark_as_representative">); That means instead of using the fields_for on line 12 of https://github.com/ndlib/sipity/blob/dltp-361/app/views/sipity/controllers/work_enrichments/attach.html.erb#L12, you would be using the <%= f.input :mark_as_representative %>
* On the Attachment model, you can add `validates :mark_as_representative, uniqueness: { scope: :work_id }`

Regarding deletion break apart the input this form object is receiving:

* New attachments are coming in via the `:files` parameter; We won't be deleting those.
* Existing attachments will be coming in via the `attachments_attributes`, some of those will have changes to file names, and some will have the _destroy value set to something "truthy" (I believe it will be '1'). As you are iterating over the `attachment_attributes`, delete the ones with _destroy

app/forms/sipity/forms/work_enrichments/attach_form.rb
```diff
@@ -6,22 +6,68 @@ class AttachForm < Forms::WorkEnrichmentForm
...
+            Array.wrap(remove_files).compact.each do |pid|
+              repository.remove_files_from(pid: pid, user: requested_by)
+            end
+            repository.mark_as_representative(pid: mark_as_representative, user: requested_by)
           end
         end
+
+        def attachments_from_work
+          return [] unless work
+          work.attachments.present? ? work.attachments : [Models::Attachment.build_default]
```
jeremyf: We don't want to build any attachments if none are associated. If we do build these attachments, then when I go to the form the first time I see a file upload field and a form for my files. I should only see a file upload field.

When I come back to the form, I should only see the existing work attachments.

So this method should look like:

```ruby
def attachments_from_work
  return [] unless work
  work.attachments
end
```

app/views/sipity/controllers/work_enrichments/attach.html.erb
```diff
@@ -4,6 +4,14 @@
...
   <%= field_set_tag do %>
     <%= f.input :files, as: :file, input_html: { multiple: 'multiple' } %>
-    <%= f.submit %>
   <% end %>
+
+  <%= field_set_tag do %>
+    <%= f.fields_for model.existing_attachments do |attachment_form| %>
+      <%= attachment_form.input :name%>
+      <%= attachment_form.input :mark_as_representative, as: :radio_buttons %>
+      <%= attachment_form.input :_destroy, boolean: true %>
```
jeremyf: When I render this form the `:_destroy` field is coming up as a text box. I believe it should be:

```ruby
<%= attachment_form.input :_destroy, as: :checkbox %>
```

db/migrate/20150205191116_add_representative_to_attachment.rb
```diff
@@ -0,0 +1,5 @@
+class AddRepresentativeToAttachment < ActiveRecord::Migration
+  def change
+    add_column :sipity_attachments, :mark_as_representative, :boolean, default: false
```
jeremyf: I don't like this as the name of the field. `:mark_as_representative` is an action you take on the item. But when we are persisting state, this is confusing.

Change from `:mark_as_representative` to `:is_representative_file`

app/models/sipity/models/attachment.rb
```diff
@@ -6,6 +6,12 @@ class Attachment < ActiveRecord::Base
       self.table_name = 'sipity_attachments'
       self.primary_key = :pid
 
+      alias_attribute :name, :file_name
+
+      def self.build_default
```
jeremyf: This is not needed.

spec/support/site_prism_support.rb
```diff
@@ -133,10 +133,20 @@ class AttachPage < SitePrism::Page
       PARAM_NAME_CONTAINER = 'work'.freeze
       element :form, "form[method='post']"
       element :input_file, "form [name='#{PARAM_NAME_CONTAINER}[files][]'][multiple='multiple']"
+      element :attachment, "form [name='#{PARAM_NAME_CONTAINER}[remove_files][]'][type='checkbox']"
```
jeremyf: I'm confused by these adjustments. What is your intent?

banurekha: I was adding elements for attachments, since I changed how attachments are rendered, this element is irrelevant.. Once the form is fully working I need to create more elements for the existing files which will be later.. So I will remove this element.

jeremyf:
@banurekha There are untested methods. https://coveralls.io/builds/1895038/source?filename=app%2Fforms%2Fsipity%2Fforms%2Fwork_enrichments%2Fattach_form.rb

jeremyf:
@banurekha There is still untested code. Please click the coverage badge and see what methods are not tested.

banurekha:
Working on them right now.

PR #137, opened 2015-02-11
DLTP-366 Adding netid validator

banurekha:
Adding validator that return true - Enhance with actual code to validate the netid
Adding spec for the validator
Include it as part of collaborator netid validation

### DIFF
--- a/app/models/sipity/models/collaborator.rb
+++ b/app/models/sipity/models/collaborator.rb
@@ -27,6 +27,7 @@ module Sipity
       #   I would prefer forms to represent business logic/validations.
       validates :name, presence: true
       validate :validate_required_information_if_responsible_for_review
+      validates :netid, net_id: true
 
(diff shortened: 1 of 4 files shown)


jeremyf:
On initialization, I would like to be able to pass in an optional collaborating lambda.

```ruby
# Make sure to match up the method signature of the EachValidator#initialize
def initialize(options = {})
  super
  @netid_remote_validator = options.fetch(:netid_remote_validator) { default_netid_remote_validator }
end
```

jeremyf:
To add some clarification. The `NetIdValidator#initialize` method is called as part of the `Collaborator.validates` class method call. It is configuring the validation rules.

When we call the instance method `Collaborator#valid?` then the `NetIdValidator#validates_each` instance method is called with the instance of the collaborator, the attribute name, and the value of that attribute.

PR #143, opened 2015-02-18
Removing work policy acting as behavior

jeremyf:
## Adding enrichment_type

@35c993a4aa12a61bb5e5e1be8d015dc58daee21e

This is a shim as I look towards splicing in the generalized behavior
for Processing queries.

## Adding ProcessingActions class

@3d0904b8bf02999cfff8c77c2b8be0b6bb68438e

This is a first step, need to tested in isolation.

## Teasing apart actions

@6d6089a00c688a11da0bcd94c8b8fb829f5f19fc

## Teasing out a StrategyAction query

@74f7925b26527e56af7a43c65ffa037a5aaa6063

> @api public
>
> An ActiveRecord::Relation scope that meets the following criteria:
>
> * Actions that are permitted to the current user
> * Actions that are available for the entity's current state.
>
> @param user [User]
> @param entity an object that can be converted into a
>   Sipity::Models::Processing::Entity
> @return ActiveRecord::Relation<Models::Processing::StrategyAction>

## Adding new actions to bootstrap

@7431a4bf264783f1e237c585eadea2b03d722416

## Adding processing actions

@ad918d2bf7d23cfccd327fb34f17ee770ced3933

## Adding EnrichmentAction class

@0f9c5e27a22f161d375f677188c797691b9152f3

Given that there are three types of actions, I want to include the
code for those actions.

## Extracting alternate action presenters

@5d041b1b691a6492d17b536e6c8b2d317e35b1ed

Moving away from the mixed bag of action rendering; There is work to
do.

## Removing configuration of view context

@e8317b0a7c3ef8ccea078010bae503d268472490

## Adding some doucmentation and removing files

@d61894583babfd371d14db561473e4c97403b3e1

## Restoring a file

@36001c63570fb20a3a995d6c16d194f8046f2c17

## Splicing in Enrichment Action view behavior

@a952c156da156f7b2c39ff4fbfc488f7639bdf90

These are the requisite changes to make that happen. There is a
failing test that I will continue to work on.

## Changing action names to match specs

@a49422d03cff77d341975148b4443e8355e59048

This also involves moving around some internationalization options.

## Removing deprecated methods

@c713a15d51c16162e6c824246f046b6a6ce6e5e8

## Removing unused processing query

@a341d457b460d77d5b5f292acae298da81ca4ffc

I thought this might be used but I'm removing it. Could be useful for
later.

## Removing unused method

@152b7d48dce693064f1ee77e3d3a5db159eb4451

## Improving code coverage

@615eee2b619a6ec74e56a24664b8948f7330aea1

This method was defined previously. Focus instead on leveraging that
abstract definition. Note: These methods are not unit tested, so I
will need to work on that.

And yes I know, not unit tested. I was driving instead from my feature
tests. My intention is to make sure to go back and add them by
assessing non-covered code via the rake tasks.

## Removing the more convoluted method declarations

@a3691ef683333c8dfed3a66e4ab952f865f92ed8

In preference for the processing subsystem.

## Removing deprecated methods

@52a4f237cc07a9eff8167fe7c9e1f6191d96f3a0

These are no longer needed as I move towards the Processing subsystem.

## Adding tests for resourceful action decorator

@6f595b0e17c0c88798cb315154f18fd8650b0dbf

These were missing in the code coverage.

## Adding specs to restore coverage

@de3c31c4ac452f5ab5a79872aad5949b61945141

This is not yet adequate as some of the units are not being tested at
the unit level but at the feature level.

## Removing unused enrichment query

@4368bd560d7ac11b355302f6965023e00b1444a6

I was keeping this around as a reference for how to do queries with
joins and AREL.


app/decorators/sipity/decorators/processing/base_decorator.rb
```diff
@@ -0,0 +1,41 @@
...
+        end
+
+        attr_reader :action, :entity
+
+        delegate :name, :action_type, to: :action
+        alias_method :label, :name
+
+        # This action, if it is rendered, is always available.
+        def availability_state
+          'available'
```
dbrower: Use ACTION_AVAILABLE here?

jeremyf:
@dbrower Done and pushed.

PR #158, opened 2015-02-24
Adding thesis defense date form

jeremyf:
## Tidying up spec and narrowing interface

@b02f88404657d176b3f24eba26c3385ee3958729

I don't want the abstract_from_work to be an exposed public method. It
is something that happens on initialization and should not be called
at a future point.

Also clarifying how the values are initialized if none are passed.

## Adding DefenseDateForm

@11993a4939d46c58f31e511b53180249ab2a05a9

Responsible for capturing information related to the defense date.

## Adding DefenseDate form and routes

@4b8460f56cd9feb094d2afc3d6fe728b865f27f8


app/views/sipity/controllers/work_enrichments/defense_date.html.erb
```diff
@@ -0,0 +1,9 @@
+<!-- TODO: Internationalize the enrichment option -->
+<h2>Tell us about your advisors</h2>
```
dbrower: Is this title correct?

PR #167, opened 2015-02-25
Fix masters thesis label

danhorst:


jeremyf:
Merging this locally, after a tweak.

PR #171, opened 2015-02-26
Renaming action for clarity

jeremyf:
Instead of the more obtuse "approve_for_ingest" make the action name
more clear by stating intent of approval not speculation on what
happens next.


rbalekai:
:+1: 
waiting for tests to pass

PR #175, opened 2015-02-26
Enforcing at least one collaborator

jeremyf:
## Adding validation for collaborators

@11d2dcfb819a4af522500680b0ecfdd86ee37266

I want to make sure that at least one of the collaborators has been
marked as "requires review"

## Marking error method as to_sentence

@908cce5a877be8d945981e5d4428be6a938d47ba

If there are multiple errors, I want them all to show up. This is how
we go about doing that.

## Removing periods from end of error messages

@ec2ee35288e0c97ca7287d09011c04ec2d00b672

Given that we are rendering sentences, I don't want to have periods in
those sentences.

## Reporting errors on base objects

@fcfa8960e178a7c7c1d164cf35c9d224e638542f

Given that I may have an error on the base object, I want a means of
reporting that information. This is the quickest way to do it. We
could craft a helper method to deal with this as well.

Or our own FormBuilder

### DIFF
--- a/config/initializers/simple_form.rb
+++ b/config/initializers/simple_form.rb
@@ -60,7 +60,7 @@ SimpleForm.setup do |config|
   # Method used to tidy up errors. Specify any Rails Array method.
   # :first lists the first message for each field.
   # Use :to_sentence to list all errors for each field.
-  # config.error_method = :first
+  config.error_method = :to_sentence
 
   # Default tag used for error notification helper.
   config.error_notification_tag = :div
--- a/config/initializers/simple_form.rb
+++ b/config/initializers/simple_form.rb
@@ -60,7 +60,7 @@ SimpleForm.setup do |config|
   # Method used to tidy up errors. Specify any Rails Array method.
   # :first lists the first message for each field.
   # Use :to_sentence to list all errors for each field.
-  config.error_method = :to_sentence
+  # config.error_method = :to_sentence
 
   # Default tag used for error notification helper.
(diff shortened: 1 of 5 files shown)


config/initializers/simple_form.rb
```diff
@@ -60,7 +60,7 @@
   # Method used to tidy up errors. Specify any Rails Array method.
   # :first lists the first message for each field.
   # Use :to_sentence to list all errors for each field.
-  # config.error_method = :first
+  config.error_method = :to_sentence
```
danhorst: I'm not sold on this change. It makes writing helpful error messages more difficult. It doesn't work well in the cases where we are verifying _both_ the presence of an option and its inclusion in a list. For example it changes the error messages for publishing_plans when creating a new work to:
> Please tell us if you intend to publish this work and Please select an option from the above list

Receiving both of these messages at the same time isn't very helpful. We should either revert this behavior or stop verifying the presence of a selection as well as it's inclusion.

jeremyf: It likely means that our conditionals for declaring validations is incorrect. I'll demonstrate when you get back.

Setting it up as the following is reasonable:

```ruby
      validates :work_type,
                presence: { message: I18n.t('sipity/forms.create_work_form.error_messages.work_type') },
                inclusion: { in: :possible_work_types, message: I18n.t('sipity/forms.error_messages.inclusion'), allow_nil: true  }
```

Relying on the order of the errors is rather arbitrary as well.

PR #181, opened 2015-02-27
Working on attachment

jeremyf:
## Adding Rack::Cache gem

@04380b4101ec964b5234a95ce9e4a6b51ddd9052

This is recommended for the Dragonfly gem, and is an all around good
idea for staging and development.

## Applying `rack_cache = true` to environments

@579803301e14a537fd78893bca56721abb0cc99a

Again, a recommendation for Dragonfly and caching in general

## Adding Dragonfly configuration

@fa1cc05190c8b35aa2d9ac472bc0dc55fa3f9d7f

@See http://markevans.github.io/dragonfly/configuration/

## Adding config/application.yml to repo

@924c3d803d856206b814cb2463e8f1c1d3099490

I want a way to share what are the names of the keys our application
is expecting. This also helps in the bootstrap process for our
application.

As a matter of deploys we update the secrets of our application, so
the keys you find within the config/application.yml are not the real
keys.

## Initial unverified stab at attachment deletion

@8dc0e4d06d24aabe4dececdaf53d0531e4dfdd8a

This is non-functioning code, but it does expose the correct
attributes for enabling deletion and rendering of existing
attachments.

## Removing validation of files

@76dd23e74dfe6612c64a937caeb8d7f85b0b8970

This is a somewhat silly requirement; Perhaps something that we can
enforce via an alternate method. What I am going for is "Is there at
least one file associated"; It can be either that I just uploaded or
it can be ones that are already uploaded.

### DIFF
--- a/config/initializers/dragonfly.rb
+++ b/config/initializers/dragonfly.rb
@@ -1,0 +1,26 @@
+require 'dragonfly'
+
+# Configure
+Dragonfly.app.configure do
+  plugin :imagemagick
+
+  secret Figaro.env.dragonfly_secret
+
+  url_format "/media/:job/:name"
+
+  datastore :file,
+    root_path: Rails.root.join('dragonfly', Rails.env),
+    server_root: Rails.root.join('public')
+end
+
+# Logger
+Dragonfly.logger = Rails.logger
+
+# Mount as middleware
+Rails.application.middleware.use Dragonfly::Middleware
+
+# Add model functionality
+if defined?(ActiveRecord::Base)
+  ActiveRecord::Base.extend Dragonfly::Model
+  ActiveRecord::Base.extend Dragonfly::Model::Validations
+end
--- a/.gitignore
+++ b/.gitignore
@@ -80,7 +80,6 @@ pickle-email-*.html
 .powenv
 
 # Ignore application configuration
-/config/application.yml
 
 # Ignore documentation
 .yardoc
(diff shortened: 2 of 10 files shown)


dbrower:
Um. Did you mean to bring in the user `See` with the @ See in the commit message?

config/initializers/dragonfly.rb
```diff
@@ -0,0 +1,26 @@
...
+# Configure
+Dragonfly.app.configure do
+  plugin :imagemagick
+
+  secret Figaro.env.dragonfly_secret
+
+  url_format "/media/:job/:name"
+
+  datastore :file,
+    root_path: Rails.root.join('dragonfly', Rails.env),
```
dbrower: This is probably not going to work in production. each capistrano deploy will get its own `dragonfly` folder.

jeremyf: Its something that will need to be symlinked as part of the capistrano task. But something to add to the punch list.

dbrower: Why not add that to this PR? If PR is merged then sipity is not going to work as expected when deployed. Just my opinion.

jeremyf: I would consider adding it to the PR, but this PR is a piece necessary for both @danhorst and I to proceed on different parts. I can then also assign the capistrano task to someone else.

I've added the concern to our punch list, so it won't be forgotten. https://docs.google.com/a/nd.edu/document/d/1Bt2o-T_h7vx3cFW7L0bVfOUci3S1xrwr1U2AHWxUWLQ/edit#

danhorst:
I'm merging with the caveat that this is WIP until the deployment issues are worked out.

PR #183, opened 2015-02-27
Working on exposing the delete action

jeremyf:
## Reworking exception to be more expressive

@89d4677af854b11748d6cd696edd86508af23500

## Guarding the resourceful action initialization

@e4b367d7d2ed17e7d80c60a4c5fd78435b255e20

As I was working on testing other things, I had a case of a poor
initialization that resulted in weird behavior. This fixes that by
throwing a noisy exception.

## Adding #render_entry_point

@14caf2eb82629603c2b3a7b5499cfdc245e97058

This method is useful for containing the logic regarding a resourceful
action. In theory, we may want to consider a confirmation action
as it relates to the destroy action; For now this will work.

## Splicing in use of render_entry_point

@2fe412635c34f0a5687b11b9b727caa87fcff28c

There is if logic that I'd like to step away from, but for now I'll be
happy to encapsulate the first part of the logic.


danhorst:
:+1:

PR #185, opened 2015-02-27
Restoring test coverage

jeremyf:
## Moving class declaration for clarity

@68f1a61ea01411b45c57d786ac031b9dcbfc0ba6

The incorrect class was at the beginning of the file. This made the
code more confusing.

## Improving clarity of intent by renaming

@2f45288fed87d24a6f09b0b9ac104013d8fb771e

Instead of the less expressive object, refering to it as an
attachment.

## Exposing a method to ease presenter logic

@ed6ac1c151a1dc2e15ac180243f84e06ed4c3e3b

This method is to help a simple wrapping element not need to be so
very nosy about its neighbor.

## Reducing the presenter's knowledge

@700713c79645ba577031b71e298e1b2914bdd213

Instead of digging deep into an attachment, I'm exposing the basic
information that I need.

## Privatizing object to narrow exposure

@d3b74b14d8769b8c619a0a3f9bcab52fe96fc1f2

This object helps in the collaboration of form building and rendering.
Its primary function is to expose a `#delete` attribute along side a
handful of methods to help render an HTML form.

### DIFF
--- a/app/forms/sipity/forms/work_enrichments/attach_form.rb
+++ b/app/forms/sipity/forms/work_enrichments/attach_form.rb
@@ -54,6 +54,7 @@ module Sipity
           attr_reader :attachment
           private :attachment
         end
+        private_constant :AttachmentFormElement
       end
     end
   end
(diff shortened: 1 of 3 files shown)


jeremyf:
Hold on this.

PR #193, opened 2015-03-02
Working on advisor requests change form

jeremyf:
## Making sure we are sending the processing comment

@27ea55263f170b9769f986337ccffc847526acb0

Instead of the entity, of which we won't know what is the processing
comment, I'm opting to send the processing comment.

## Removing trailing blank space

@ef3342fdf03914e071b1f4a1e67af89a6328b43a

[skip ci]

## Removing un-needed module declration

@f4aaed26b38e081e92b8474596a85e42ab51f36b

## Extracting the AdvisorRequestsChangeDecorator

@6ac495b4c1dd52aa7f39c26bad6c4b6a1870169d

This is a requirement as we want to render a comment specific email.
So we will extract that information from the processing comment and
present it as appropriate.

## Refactoring #advisor_requests_change

@322b879974e2e6dc1daa85dc86dca9285eaed0ab

The one-size fits all decorator does not fit (but was helpful to get
us addressing the next important updates).

This is splicing in the new Email::AdvisorRequestsChangeDecorator.

## Fleshing out AdvisorRequestsChangesDecorator

@cf2fecf49f8bccfe4a8d71df46d851365c02c89b

Wiring in the required retrieval information

### DIFF
--- a/app/decorators/sipity/decorators/emails/advisor_requests_change_decorator.rb
+++ b/app/decorators/sipity/decorators/emails/advisor_requests_change_decorator.rb
@@ -1,18 +1,22 @@
 module Sipity
   module Decorators
     module Emails
+      # Responsible for exposing methods required for email delivery. Nothing
+      # too fancy.
       class AdvisorRequestsChangeDecorator
         def initialize(processing_comment)
           @processing_comment = processing_comment
         end
 
         def name_of_commentor
+          actor.proxy_for.name
         end
 
         def document_type
+          entity.proxy_for.work_type
         end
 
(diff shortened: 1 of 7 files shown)


banurekha:
So the idea is to have separate email decorator for each notifier method? so we get required value from the notifier?

jeremyf:
@banurekha Perhaps; we won't know until we look at each email in isolation. Some of the emails are related to a comment. Others are related to the entity as a whole. At this point, I believe what @rbalekai and I put in place the pattern.

PR #195, opened 2015-03-02
Add account profile

banurekha:


jeremyf:
I don't like the commits with tests show up after the commit with production code. Please squash these commits so they are one concept. Then I'll review.

PR #206, opened 2015-03-04
Working on email decorators

banurekha:


banurekha:
These values are used in email at the moment and I had preferred them in decorator since there is no styling needed at email.

app/decorators/sipity/decorators/email_notification_decorator.rb
```diff
@@ -42,10 +36,21 @@ def work_show_path
         view_context.work_path(entity.id)
```
jeremyf: We can't use `work_path`; That resolves to `/works/:id`. We need the URL, so we have `http://sipity-staging.library.nd.edu/works/:id`

I believe we will need a configuration option.

jeremyf:
@banurekha Sounds reasonable. Keep the `to_sentence` usage.

jeremyf:
I'm rebasing and going to merge. Good stuff!

PR #210, opened 2015-03-05
Working on ldap client

jeremyf:
## Reworking LDAP client for readability

@ccf8847613162fc0a81e4f336e2a0f2867393dcb

The specs were in the wrong file, and thus need to be fired.

## Changing LdapClient initialization options

@11a6aeed3a0743ee002d51f672e2bbebec62397c

## Reworking LdapClient to ensure coverage

@a9b2ce9400977ff3f322583a7a0b69caf88a7c85

The client was confusing. There were two methods that queried.

One was a method that caught a time out exception. It would then raise
a UserNotFound error. That is confusing, because the exception is now
translating the timeout error to something with a different meaning.

So I've opted to remove that; as it wasn't called. Also, I'm factoring
towards accepting an connection instead of instantiating a new one;
Perhaps there should be a connection that is initialized as part of
boot process of the application. Don't know and not yet important.

## Exposing convenience method & Figaro interaction

@0addcff26d37bd68073918ae8d580911ddd9b45f

## Splicing in LdapClient validator

@33265050cbaf50075e88fbc360539c458035179e


app/services/sipity/services/ldap_client.rb
```diff
@@ -3,49 +3,37 @@ module Sipity
...
     # This class querries LDAP to get user information
     class LdapClient
-      attr_reader :net_id, :connection
+      def self.valid_netid?(netid)
+        configuration = {
+          host: Figaro.env.ldap_host,
+          port: Figaro.env.ldap_port,
+          encryption: Figaro.env.ldap_encryption.to_sym
+        }
+        new(configuration).valid_netid?(netid)
```
dbrower: Since the same (?) configuration is always pulled from Figaro, does it make sense to cache the object created here with `new` in a class variable?

rbalekai:
:+1:

app/services/sipity/services/ldap_client.rb
jeremyf: My intention is to extract the connection object into something that gets passed into this class. However that is a future consideration.

dbrower: ok. no problem.

Does passing in the connection mean people using the class need to know this class uses NET::LDAP as the client library?

PR #212, opened 2015-03-05
Working on ability to delete collaborators

jeremyf:
## Adding method for managing collaborators

@ded05e7d54ec0c975e873be997635cc09ff93ec2

> Responsible for adding collaborators to the work and removing anyone
> else.

## Adding ability to remove collaborators

@fef6d891a8f8cdd551bebca2328c01e69a0b6b6a

Since we have the repeater field, I'm making an assumption. If the
user removes the collaborator via the remove button in the UI, then
I will delete that collaborator.

## Destroying collaborator destroys its actor

@1745a1df4782c8d1658805a0eeb154a5c05585d3

Unlike a user and a group, a collaborator-based actor is much more
transient; That is to say Sipity is exposing a user based means of
getting rid of collaborators.

## Pushing collaborator build logic into form

@99bdd817b4fbbea803a36ebf8cc738dc5c635048

As per @danhorst's TODO items, I'm moving the building an extra
collaborator logic down into the form.

Also, don't worry about stripping away extract empty rows from the UI.
It is being [handled here](https://github.com/ndlib/sipity/blob/e75508b1566624133f72f021bde439dba3f5ddc7/app/forms/sipity/forms/work_enrichments/collaborator_form.rb#L56).
See the guard condition of the below method.

```ruby
def build_collaborator_from_input(collection, attributes)
  return if !attributes[:name].present? && !attributes[:id].present?
  collaborator = repository.find_or_initialize_collaborators_by(work: work, id: attributes[:id])
  collaborator.attributes = extract_collaborator_attributes(attributes)
  collection << collaborator
end
```

## Restoring collaborator errors on base

@f007d8fea94a73bd0b4e2586c5ee252ea946cb45

This is an ugly solution, but one to revisit. The validations of the
form throw an error on base if NO collaborators are given that require
review. That error message does belong with any of the collaborators.

It would be possible to push that error onto each of the items. But
we'd need to revisit how we'd do that.

## Gathering validation methods to common location

@23c226c20fe8085bad51774cfc89786cb563bc5e

Not only am I moving the methods, but I'm also wrapping the possible
input because it could be nil.

That is one concession concerning the contributors, I won't set a user
input if I receive an empty hash. But validation should be against
the collaborator attributes that were provided.

### DIFF
--- a/app/forms/sipity/forms/work_enrichments/collaborator_form.rb
+++ b/app/forms/sipity/forms/work_enrichments/collaborator_form.rb
@@ -51,7 +51,9 @@ module Sipity
 
         def collaborators_from_work
           return [] unless work
-          work.collaborators.present? ? work.collaborators : [Models::Collaborator.build_default]
+          # Manually building an empty collaborator to allow adding more once
+          # one is already created:
+          work.collaborators + [Models::Collaborator.build_default]
         end
 
         def each_collaborator_must_be_valid
--- a/app/forms/sipity/forms/work_enrichments/collaborator_form.rb
+++ b/app/forms/sipity/forms/work_enrichments/collaborator_form.rb
@@ -7,15 +7,16 @@ module Sipity
           super
           self.collaborators_attributes = attributes[:collaborators_attributes]
         end
-        attr_reader :collaborators_attributes
+        attr_reader :collaborators_attributes, :collaborators_from_input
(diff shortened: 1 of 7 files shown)


rbalekai:
:+1:

danhorst:
:+1:  (better late than never)

PR #216, opened 2015-03-05
Adding meta_request gem

jeremyf:
Paired with the [RailsPanel](https://github.com/dejan/rails_panel)
Chrome Extension, I can get lots of information related to rendering
and queries.

> The first columns gives you the HTTP Response Status, the Controller
> and Action that your route hit, the HTTP Verb (PUT/POST/GET),
> request format (HTML, JSON, XML) and the response time. These
> metrics by itself would have been quite valuable, but there is more.
> You can click on any request that you made and see more information
> about it.
>
> ...
>
> The next tab is the Rendering which gives a breakdown of all the
> views being rendered. This is useful for finding which partials are
> taking the longest to render. Click on the view name automatically
> opens it in your editor.

Discovered from the following blog post:
http://www.rubyonrails365.com/7-must-have-gems-to-install-on-any-project/

Consider adding the newrelic_rpm gem for local reporting.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,6 +56,9 @@ group :development do
   gem 'pry-byebug'
   gem 'letter_opener'
   gem 'seed_dump'
+  # Paired with Chrome the RailsPanel plugin, you can see request information
+  # https://github.com/dejan/rails_panel
+  gem 'meta_request'
 end
 
 group :development, :test do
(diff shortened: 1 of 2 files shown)


jeremyf:
Please review the usage of the gem that I'm adding: @danhorst @rbalekai @banurekha @dbrower

Once reviewed give a :+1: and if you are the last then feel free to merge.

rbalekai:
:+1:

danhorst:
This looks great! :heavy_plus_sign::100:

dbrower:
Looks good. It is only for the development environment, so +1.

banurekha:
I like it too.

PR #217, opened 2015-03-06
Additional attribute degree

banurekha:


app/forms/sipity/forms/work_enrichments/degree_form.rb
```diff
@@ -0,0 +1,33 @@
...
+        private
+
+        def save(requested_by:)
+          super do
+            repository.update_work_attribute_values!(work: work, key: 'degree', values: degree)
+          end
+        end
+
+        def degree_from_work
+          Queries::AdditionalAttributeQueries.work_attribute_values_for(work: work, key: 'degree')
```
jeremyf: use `repository.work_attribute_values_for(work: work, key: 'degree')` instead of the module function.

app/repositories/sipity/queries/simple_controlled_vocabulary_queries.rb
```diff
@@ -0,0 +1,12 @@
+module Sipity
+  module Queries
+    # Queries
+    module SimpleControlledVocabularyQueries
+      def get_values_by_predicate_name(name:)
+        Models::SimpleControlledVocabulary.where(predicate_name: name).pluck(:predicate_value)
+      end
+      module_function :get_values_by_predicate_name
```
jeremyf: remove `module_function :get_values_by_predicate_name` and `public :get_values_by_predicate_name`

spec/forms/sipity/forms/work_enrichments/degree_form_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+          it 'will have #degree_names' do
+            allow(repository).to receive(:get_values_by_predicate_name).and_return(name)
+            expect(subject.degree_names).to_not be_empty
+          end
+
+          context 'with data from the database' do
+            let(:degree_name) { 'test' }
+            subject { described_class.new(work: work, degree: degree_name, repository: repository) }
+            it 'will return the degree of the work' do
+              allow(subject).to receive(:degree_from_work).and_return(name)
```
jeremyf: switch to

```ruby
allow(repository).to receive(:get_values_by_predicate_name).with('degree').and_return([names])
```

Because the underlying `pluck` call will return an array.

jeremyf:
Still one more. Replace the following:

```ruby
allow(subject).to receive(:degree_from_work).and_return(name)
```

with

```ruby
allow(repository).to receive(:work_attribute_values_for).with(work: work, key: 'degree').and_return([names])
```

The reason is that I want to make sure that the repository method is being called. Also, the underlying repository method returns an array, which is different from what you are stubbing.

PR #241, opened 2015-03-09
Working on exception handling

jeremyf:
## Removing unused exceptions

@9f6af8ae0e165dcc39adf73da9046e07a6c36941

These are exceptions that were used in previous iterations but are no
longer relevant.

## Configuration :unauthorized exceptions

@a4e0e0c8b6e0b34e11412d395066753d2a866b10

> * config.action_dispatch.rescue_responses configures what exceptions
>   are assigned to an HTTP status. It accepts a hash and you can
>   specify pairs of exception/status.
>
> From http://guides.rubyonrails.org/configuring.html#configuring-action-dispatch

## Adding public/401.html page

@6dcf0bd342cb9c7ac5ef8934e3a329bb43e7bca8

Instead of a crazy 500 exception error, I'm opting to provide an
unauthorized page 401 page. This is inadequate, but something we can
proceed with in a meaningful way.

### DIFF
--- a/app/exceptions/sipity/exceptions.rb
+++ b/app/exceptions/sipity/exceptions.rb
@@ -9,11 +9,6 @@ module Sipity
     class RuntimeError < ::RuntimeError
     end
 
-    # When something about the entity for a given policy is just not quite
-    # right.
-    class PolicyEntityExpectationError < RuntimeError
-    end
-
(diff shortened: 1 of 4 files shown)


danhorst:
:+1:

PR #288, opened 2015-03-10
Working on graph generation in dot format

jeremyf:
I want to see what are the current actions available, who can take
them and how the states transition.

[skip ci]


danhorst:
:+1: The output looks good to me.

PR #312, opened 2015-03-13
Issue 256 - Pid minter service

### LINKED ISSUE #256: Incoroporate PID Minter for Production and Pre-Production
At present, the PID minter is defined:

```ruby
def default_pid_minter
  -> { SecureRandom.urlsafe_base64(nil, true) }
end
```

[See source](https://github.com/ndlib/sipity/blob/7d31891709e84ab5a319a756542c0addf2233ee3/app/repositories/sipity/commands/work_commands.rb#L117-L120)

Instead the PID minter should be configured in the application and can be overridden in each environment, much like [#default_netid_remote_validator ](https://github.com/ndlib/sipity/blob/7d31891709e84ab5a319a756542c0addf2233ee3/config/application.rb#L26-L29)

banurekha:
closes #256

### DIFF
--- a/app/services/sipity/services/noid_minter.rb
+++ b/app/services/sipity/services/noid_minter.rb
@@ -1,27 +1,40 @@
 module Sipity
   module Services
+    # This class uses noid client to mint a pid
     class NoidMinter
+      def self.call(configuration = {})
+        new(configuration: configuration).call
+      end
+
       def initialize(configuration)
-        configuration = {
-            host: Figaro.env.noid_server!,
-            port: Figaro.env.noid_port!
-        }
-        new(configuration).call
+        @server = configuration.fetch(:server) { default_server }
+        @port = configuration.fetch(:port) { default_port }
+        @pool = configuration.fetch(:pool) { default_pool }
       end
 
-      # The ldap_options is a hash containing
-      # :host, :port, :encryption and :base
-      def initialize(noid_options)
-        @connection = ::NoidsClient::Connection.new(noid_options.fetch[:server])
-                      .get_pool(noid_options.fetch[:pool])
-      end
+      attr_reader :connection
 
-      def template
-        @template ||= connection.template.split("+").first
+      def connection
+        @connection = ::NoidsClient::Connection.new("#{@server}:#{@port}").get_pool(@pool)
       end
 
       # Returns a single NOID
       def call
+        connection.mint.first
+      end
+
+      private
+
+      def default_server
+        Rails.application.secrets.noid_server
+      end
+
+      def default_port
+        Rails.application.secrets.noid_port
+      end
+
+      def default_pool
+        Rails.application.secrets.noid_pool
       end
     end
   end
--- a/app/services/sipity/services/noid_minter.rb
+++ b/app/services/sipity/services/noid_minter.rb
@@ -1,0 +1,17 @@
+module Sipity
+  module Services
+    class NoidMinter
+      def initialize(configuration)
+        configuration = {
+            host: Figaro.env.noid_server!,
+            port: Figaro.env.noid_port!
+        }
+        @service ||= ::NoidsClient::Connection.new(@server).get_pool(@pool)
+      end
+
+      # Returns a single NOID
+      def call
+      end
+    end
+  end
+end
--- a/app/services/sipity/services/noid_minter.rb
+++ b/app/services/sipity/services/noid_minter.rb
@@ -2,29 +2,19 @@ module Sipity
   module Services
     # This module uses noid client to mint a pid
     module NoidMinter
-      def self.call(configuration = {})
-        mint_a_pid(configuration: configuration)
-      end
+      module_function
 
-      def mint_a_pid(configuration)
-        @server = configuration.fetch(:server) { Figaro.env.noid_server! }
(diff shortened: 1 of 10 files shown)


app/services/sipity/services/noid_minter.rb
```diff
@@ -0,0 +1,41 @@
...
+        new(configuration: configuration).call
+      end
+
+      def initialize(configuration)
+        @server = configuration.fetch(:server) { default_server }
+        @port = configuration.fetch(:port) { default_port }
+        @pool = configuration.fetch(:pool) { default_pool }
+      end
+
+      attr_reader :connection
```
jeremyf: Unneeded method; the `def connection` below obliterates this.

app/services/sipity/services/noid_minter.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      def connection
```
jeremyf: Make `connection` a private method (move it just above `default_server`)

spec/services/sipity/services/noid_minter_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        { server: 'noid.example.com', port: '13001', pool: "pool_name" }
+      end
+      subject { described_class.new(configuration) }
+
+      it 'exposes .call as a convenience method' do
+        expect_any_instance_of(described_class).to receive(:call)
+        described_class.call(configuration)
+      end
+
+      context 'connection' do
```
jeremyf: Don't need to test this as you are making it a private method.

spec/services/sipity/services/noid_minter_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        end
+      end
+
+      context "#call" do
+        subject { described_class.new(configuration) }
+        let(:pid) { 'a pid' }
+        context 'with valid options' do
+          it "will return a pid minted by noid server" do
+            expect(subject).to receive(:connection).and_return(connection)
+            expect(connection).to receive(:mint).and_return([pid])
```
jeremyf: The above two lines are avoiding the Noid client. Lets make sure the configuration is working. And then override mint.

```ruby
expect(subject.send(:connection)).to receive(:mint).and_return([pid])
```

app/services/sipity/services/noid_minter.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      # Returns a single NOID
+      def call
+        connection.mint.first
+      end
+
+      private
+
+      def default_server
+        Rails.application.secrets.noid_server
```
jeremyf: Move to Figaro, and make it required via the initializer.

app/services/sipity/services/noid_minter.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      def default_port
+        Rails.application.secrets.noid_port
```
jeremyf: Move to Figaro, and make it required via the initializer.

app/services/sipity/services/noid_minter.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      def default_pool
+        Rails.application.secrets.noid_pool
```
jeremyf: Move to Figaro, and make it required via the initializer.

app/services/sipity/services/noid_minter.rb
```diff
@@ -0,0 +1,41 @@
...
+      def initialize(configuration)
+        @server = configuration.fetch(:server) { default_server }
+        @port = configuration.fetch(:port) { default_port }
+        @pool = configuration.fetch(:pool) { default_pool }
+      end
+
+      attr_reader :connection
+
+      def connection
+        @connection = ::NoidsClient::Connection.new("#{@server}:#{@port}").get_pool(@pool)
```
dbrower: use `||=` here?

config/application.rb
```diff
@@ -28,6 +28,10 @@ class Application < Rails::Application
       true
     end
 
+    config.default_pid_minter = lambda do
+      Services::NoidMinter.call
+    end
```
dbrower: by using the noid minter class in this way, the connection to the noid server is not being cached (since the caching is done in each instance of the class, but calling the class itself, the code is making a new instance each time). Is there a reason to not cache the connection? Each connection requires two API calls to get set up. So single identifier minted will make 3 API calls and take up about 8 ms of time.

jeremyf:
Get this test to pass.

```ruby
it 'will have a cached connection' do
  first_connection = described_class.connection
  next_connection = described_class.connection
  expect(first_connection.object_id).to eq(next_connection.object_id)
end
```

jeremyf:
This should be adequate.

```ruby
module Sipity
  module Services
    # This module uses noid client to mint a pid
    module NoidMinter
      module_function

      def call
        connection.mint.first
      end

      def connection
        return @connection if @connection.present?
        server = Figaro.env.noid_server!
        port = Figaro.env.noid_port!
        pool = Figaro.env.noid_pool!
        @connection = ::NoidsClient::Connection.new("#{server}:#{port}").get_pool(pool)
      end
    end
  end
end
```

PR #330, opened 2015-03-16
locking 'sanitize' gem to 2.1.0 to have it working

rbalekai:
updating sanitize to 3.0.0 versions breaks on my machine


danhorst:
:+1:  I would like to see [gumbo](https://github.com/google/gumbo-parser) integration but that can wait.

PR #347, opened 2015-03-17
Removing with action pane

jeremyf:
## Removing #with_action_pane

@30b603d1b705b307a257e06c7eb6e6016f556468

This was a kludge and resulted in decorating an object within the
repository methods; Something that I want to avoid. So I'm attempting
to streamline the behavior.

## Removing unneeding dom classifications

@8656e5c58d40668b460c2a5b583ad8a78cbc163d

These extra DOM classes were chatty and by appearance nothing was
contingent on their presence.


danhorst:
:+1:

PR #355, opened 2015-03-17
Display comments

### LINKED ISSUE #313: Render “feedback” comments on work show page
When an advisor, or the graduate school reviewer, leaves a comment it needs to be rendered on the page so that the depositor can see it. (We’re assuming that simply including the comment in an email is insufficient to communicate the feedback.)

rbalekai:
closes #313

### DIFF
--- a/app/decorators/sipity/decorators/work_decorator.rb
+++ b/app/decorators/sipity/decorators/work_decorator.rb
@@ -57,6 +57,12 @@ module Sipity
         end
       end
 
+      def comments
+        @comments ||= repository.find_comments_for_work(work: self).map do|comment|
+          Decorators::Processing::ProcessingCommentDecorator.decorate(comment)
+        end
+      end
+
       private
 
       def processing_actions(user:)
--- a/app/repositories/sipity/queries/comment_queries.rb
+++ b/app/repositories/sipity/queries/comment_queries.rb
@@ -1,0 +1,37 @@
+module Sipity
+  module Queries
+    # Queries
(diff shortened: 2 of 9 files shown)


app/decorators/sipity/decorators/work_decorator.rb
```diff
@@ -57,6 +57,12 @@ def enrichment_actions(user:)
         end
       end
 
+      def comments
+        @processed_comments ||= repository.find_comments_for_work(work: self).map do|comment|
```
jeremyf: Where is the instance variable different from the method name? By default, a memoized (i.e. `||=`) method should have an instance variable that is the same as the method name.

```ruby
def comments
  @comments ||= ...
end
```

app/repositories/sipity/queries/comment_queries.rb
```diff
@@ -0,0 +1,11 @@
+module Sipity
+  module Queries
+    # Queries
+    module CommentQueries
+      def find_comments_for_work(work:)
+        entity = Conversions::ConvertToProcessingEntity.call(work)
+        Sipity::Models::Processing::Comment.where(entity_id: entity)
```
jeremyf: Even though the above works, replace `(entity_id: entity)` with `(entity_id: entity.id)`; Its more clear and consistent.

jeremyf: Also, where is the corresponding spec? Should the comments be returned in a specific order?

Issue #360, opened 2015-03-18
What should be the order of the comments?


rbalekai:
Should it be in descending order of 'created_at' field?

danhorst:
List comments chronologically, newest first

PR #361, opened 2015-03-18
Display comments

### LINKED ISSUE #313: Render “feedback” comments on work show page

rbalekai:
closes #313 

related to #360


jeremyf:
:+1: Waiting on Travis!

Issue #363, opened 2015-03-18
Be able to list of all comments given on a work


danhorst:
List them chronologically in descending order.
There will need to be an index on the `created_at` date for comments.

PR #371, opened 2015-03-19
Comment order

rbalekai:


app/views/sipity/controllers/works/show.html.erb
```diff
@@ -73,17 +73,19 @@
     </div>
   </div>
 
-  <% if model.comments.present? %>
+  <% if model.latest_comment.respond_to?(:comment) %>
```
jeremyf: Revert the changes to this file; They are not "in scope" for #363.

app/repositories/sipity/queries/comment_queries.rb
```diff
@@ -4,7 +4,11 @@ module Queries
     module CommentQueries
       def find_comments_for_work(work:)
         entity = Conversions::ConvertToProcessingEntity.call(work)
-        Sipity::Models::Processing::Comment.where(entity_id: entity.id)
+        Sipity::Models::Processing::Comment.where(entity_id: entity.id).order('created_at DESC')
+      end
+
+      def find_latest_comment(work:)
```
jeremyf: We don't need `find_latest_comment`; That is for a different issue.

Remove the change from the commit. I believe the method you are after will continue to retrieve an array of comments; `find_comments_related_to_preceding_state`

PR #372, opened 2015-03-20
Articulating expectations regarding code

jeremyf:
These are some of the expectations that I have internalized, but I
want to state them clearly and articulate why I believe they should be
part of the policy regarding our build.

[skip ci]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ Therefore lets keep it under 30 seconds.
 
 Code that is not covered in test means code that we are not "owning" and documenting; Therefore lets keep the test coverage at 100%.
 
-Code that does not conform to the style guide can be disorienting; Therefore lets adhear to 100% style guide compliance or articulate why we are violating the styleguide.
+Code that does not conform to the style guide can be disorienting; Therefore lets adhear to 100% style guide compliance or articulate why we are violating the style guide.
 
 ### What Happens We Pass the 30 Second Threshold?
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,18 +47,25 @@ SPEC_OPTS="--format progress" bin/rake spec:travis' > .git/hooks/pre-push ; chmo
 
 If you need to skip the `pre-push` hook, use the `--no-verify` option when you push to a branch.
 
-## Build Requirements/Policies
+## Build Goals
 
 * Our test suite must complete in 30 seconds or less
 * Our code coverage must be 100%
 * Our code must have 100% style guide compliance
   - Or a clear reason for exclusion
+* Our code will have a 3.75 or higher grade on [Code Climate](https://codeclimate.com/github/ndlib/sipity)
 
-A slow test suite is an indicator of an unhealthy test suite; Therefore lets keep it fast.
+A slow test suite is an indicator of an unhealthy test suite;
+Therefore lets keep it under 30 seconds.
 
-Code that is not covered in test means code that we are not "owning" and documenting; Therefore keep test coverage high.
+Code that is not covered in test means code that we are not "owning" and documenting; Therefore lets keep the test coverage at 100%.
 
-Code that does not conform to the style guide can be disorienting; Therefore lets adhear to the style guide.
+Code that does not conform to the style guide can be disorienting; Therefore lets adhear to 100% style guide compliance or articulate why we are violating the styleguide.
+
+### What Happens We Pass the 30 Second Threshold?
+
+Our application is going to continue to grow; And so will the completion time of our tests.
+If we hit the 30 second mark, we'll review what is happening and adjust the goal.
 
 ## But Rubocop is complaining too much...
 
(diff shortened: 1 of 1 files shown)


jeremyf:
For your review and comments @danhorst @banurekha @rbalekai @dbrower 

I am also going to articulate that I envision myself as the Code Steward for this project. I am in charge of its architectural vision and am its Editor in Chief if you will. Again, stating clear expectations. And any of those expectations are not set in stone, but instead are a conversation.

CONTRIBUTING.md
```diff
@@ -47,6 +47,19 @@ SPEC_OPTS="--format progress" bin/rake spec:travis' > .git/hooks/pre-push ; chmo
...
 If you need to skip the `pre-push` hook, use the `--no-verify` option when you push to a branch.
 
+## Build Requirements/Policies
+
+* Our test suite must complete in 30 seconds or less
+* Our code coverage must be 100%
+* Our code must have 100% style guide compliance
+  - Or a clear reason for exclusion
+
+A slow test suite is an indicator of an unhealthy test suite; Therefore lets keep it fast.
```
dbrower: 1. I would expect the size of the test suite to grow polynomially in the size of the production code (especially if you want 100% test coverage). Hence a slow test suite may only be a sign of a large code base, and not necessarily a unhealthy test suite.
1. Should the test suite be fast, or should it run in <= 30 seconds. This is ambiguous. Choose one.
1. This line seems to be more a philosophy than a requirement. Maybe just delete it, or move it to a different section.

CONTRIBUTING.md
```diff
@@ -47,6 +47,19 @@ SPEC_OPTS="--format progress" bin/rake spec:travis' > .git/hooks/pre-push ; chmo
...
+
+Code that is not covered in test means code that we are not "owning" and documenting; Therefore keep test coverage high.
```
dbrower: ditto to above comment. high test coverage is not the same as 100% test coverage, and this is more a philosophy than a requirement.

CONTRIBUTING.md
```diff
@@ -47,6 +47,19 @@ SPEC_OPTS="--format progress" bin/rake spec:travis' > .git/hooks/pre-push ; chmo
 
 If you need to skip the `pre-push` hook, use the `--no-verify` option when you push to a branch.
 
+## Build Requirements/Policies
+
+* Our test suite must complete in 30 seconds or less
+* Our code coverage must be 100%
```
dbrower: Why not start at a more manageable goal, say 90% coverage.

Also, in my opinion, there is a big difference between 99% and 100%. perfection is generally not worth it.

dbrower:
I agree with the idea of having a single person own each repository. And I second your self-nomination to be the architect and owner of the sipity code base. :)

banurekha:
@jeremyf I tend to agree with @dbrower observation.. When code base grows it is more than likely that test suite takes more than 30 seconds and may not necessarily mean slow test suite.. Especially when you start adding view test. Also IMHO it is hard to achieve 100% and it is a very aggressive goal. Again thanks for taking this initiative.

CONTRIBUTING.md
jeremyf: We have 100% test coverage right now. I believe it is of minimal cost to maintain that.

CONTRIBUTING.md
jeremyf: I agree that it is a philosophy, but it is one that numerous authors (Kent Beck, Robert Martin, Michael Feathers, etc.) advocate as a requirement. Not because it means the code works, but it means that we are at least documenting each and every line of production code.

CONTRIBUTING.md
jeremyf: I'm articulating, in an update, what happens when our test suite goes beyond 30 seconds. The answer is that we review the policy and likely bump the number. But it is a threshold for review.

Curate had moments in time in which the test suite time doubled when a few ideas were introduced. If anything, by stating a number, it is more of a goal.

I'm going to rename from Build Requirements to Build Goals; To convey the philosophical intent.

jeremyf:
@banurekha A good observation. We already have 100% code coverage. So staying vigilant is much less expensive than if we were playing catchup.

From memory there have been 3 pull requests that have dropped the code coverage: one from me, one from @rbalekai and one from you

In previous cases I've accepted the change then went in and restored the code coverage. I'm very willing to continue helping to do that.

jeremyf:
And to be very clear, this tooling is to make sure we understand our team's capacity and output. It is not for monitoring or performance metrics of individuals. And I will do whatever is in my power to ensure that these metrics are never used for that. I also believe that is true of @danhorst will agree with my sentiment.

danhorst:
:+1: I'll back Jeremy up on this one. 100% code coverage is feasible for _this project_ because of the existing test coverage and because of the tooling that is already in place.

I appreciate that Jeremy has been taking the time to add tests to merged pull requests after the fact but that extra effort is not sustainable. If we are going to maintain complete coverage it has to be a shared commitment.

dbrower:
I like this revision since they are build _goals_ and not requirements.

rbalekai:
:+1:

banurekha:
:+1:

PR #378, opened 2015-03-23
351 issue on GitHub

rbalekai:
I18n for file access permission error


danhorst:
:+1:  I like this solution

PR #401, opened 2015-03-27
223 issue on GitHub

rbalekai:
Multi-field validator


jeremyf:
Scenarios

```ruby
[
  values: [], valid?: false,
  values: [''], valid?: false,
  values: nil, valid?: false,
  values: '', valid?: false,
  values: ['work'], valid?: true,
  values: ['work', ''], valid?: false,
  values: ['work', nil], valid?: false,
].each_with_index do |example, index|
  it "will validate #{example[:values].inspect} as #{example[:valid?].inspect} (Scenario ##{index})" do
    # do some code.
  end
end
```

jeremyf:
I'm closing this pull request at this point.

PR #418, opened 2015-03-31
417: Clarify the process for finding work to do

danhorst:
Formalizing process can be a bit pedantic but I would rather be explicit then leave room for confusion.
What do you think @ndlib/dlt?

[Here is the document as it will be look when merged](https://github.com/ndlib/sipity/blob/417-issue-on-github/CONTRIBUTING.md).

This pull request should not be merged until everyone has weighed in.


jeremyf:
:+1: The key thing to capture is that we pick from the prioritized list. If we don't know how to act on the prioritized ticket, then seek clarification. In the meantime you can go and find another ticket that might be small in scope.

dbrower:
This looks fine to me.

To be pedantic, I don't like the definition of an issue since:

 * all issues are actionable
 * a question is an issue with a special tag which is not necessarily actionable
 * so a question issue may not be actionable
 * contradiction since all issues are actionable.

I guess this is the main thing that stuck out to me.

jeremyf:
@banurekha @rbalekai Any thoughts? :+1: or :-1: or suggestions?

rbalekai:
:+1:

banurekha:
:+1:

cdeluca495:
Forgive me for asking, but if it is decided that we want to do dev on VM environments instead of local machines, how much of this process will change and what would it look like?

jeremyf:
@cdeluca495 Little of this will change if we do dev on a VM vs. local machine.

jeremyf:
@danhorst Do you have clarifications so we can get this critter reviewed and merged?

cdeluca495:
:+1:

PR #458, opened 2015-04-06
451 create a method class that fetches a given netid s preferred name

banurekha:


app/services/sipity/services/netid_query_service.rb
```diff
@@ -0,0 +1,60 @@
...
+        person.fetch('netid')
+      rescue KeyError
+        false
+      end
+
+      private
+
+      def person
+        parse
+      rescue OpenURI::HTTPError
```
jeremyf: This `rescue` appears to be out of place. I'd expect it to be closer to the location in code where the exception is thrown.

app/services/sipity/services/netid_query_service.rb
```diff
@@ -0,0 +1,60 @@
...
+        {}
+      end
+
+      def response
+        # Leveraging 'open-uri' and its easy to use interface
+        open(url).read
+      end
+
+      def parse
+        JSON.parse(response).fetch('people').first
```
jeremyf:
```ruby
def parse
  # Leveraging 'open-uri' and its easy to use interface
  response = open(url).read
  JSON.parse(response).fetch('people').first
rescue OpenURI::HTTPError
  {}
end
```

PR #460, opened 2015-04-07
412 issue on GitHub

rbalekai:
@jeremyf review?


jeremyf:
It looks like it is time to introduce an attribute presenter; If logic in the view is a key indicator that something needs to be teased apart.

jeremyf:
If you would like, we can craft a "Technical Debt" issue saying "Create an AttributeDecorator to handle custom predicate rendering" then tidy up your pull request.

PR #462, opened 2015-04-07
337 email refinements i18n for access controls embargo date persistence research director plurality

rbalekai:


jeremyf:
If we go ahead and create an issue for "Create an AttributePresenter to handle custom predicate rendering" then this pull request is closer to completion.

PR #464, opened 2015-04-09
461 app crash while updating access policies

rbalekai:


jeremyf:
I'm going to need a test for this.

PR #490, opened 2015-04-14
309 allow rich text for some attributes

banurekha:


app/views/sipity/controllers/dashboards/index.html.erb
```diff
@@ -33,7 +33,7 @@
       <tbody>
         <% view.works.each do |work| %>
           <tr>
-            <td><%= link_to(work.title, work_path(work)) %></td>
+            <td><%= link_to(raw(work.title), work_path(work)) %></td>
```
jeremyf: We can't use `raw` as this is a user input. I believe what we need is to sanitize the `title` input via a tag-whitelist then output the title as html_safe.

In using `raw` we expose a vector for anyone to inject malicious javascript that could do lots of bad things.

@danhorst do you have some suggestions.

app/views/sipity/controllers/dashboards/index.html.erb
```diff
@@ -33,7 +33,7 @@
       <tbody>
         <% view.works.each do |work| %>
           <tr>
-            <td><%= link_to(work.title, work_path(work)) %></td>
+            <td><%= link_to(work.to_s, work_path(work)) %></td>
```
jeremyf: Why not `work.title`?

app/views/sipity/controllers/works/show.html.erb
```diff
@@ -64,7 +64,7 @@
   <div class="row">
     <div class="col-xs-12 work-show page-header">
       <h1 class="work-title">
-        <span itemprop="name"><%= model.title %></span>
+        <span itemprop="name"><%= model.sanitize_html(model.title) %></span>
```
jeremyf: Can the presenter instead auto-sanitize the output of title?

app/forms/sipity/forms/work_enrichments/describe_form.rb
```diff
@@ -26,6 +26,23 @@ def save(requested_by:)
           end
         end
 
+        include Conversions::SanitizeHtml
+        def sanitize_html(value)
+          Conversions::SanitizeHtml.call(value)
```
jeremyf: If you are using the `include Conversions::SanitizeHtml` you don't need to define the `sanitize_html` method; Its included. So I believe lines 30 to 32 can be removed.

app/decorators/sipity/decorators/work_decorator.rb
```diff
@@ -39,9 +39,9 @@ def accessible_objects
         @accessible_objects ||= repository.access_rights_for_accessible_objects_of(work: object)
       end
 
-      include Conversions::ConvertToRichText
-      def rich_text_value(attribute)
-        Conversions::ConvertToRichText.call(attribute).html_safe
+      include Conversions::SanitizeHtml
+      def sanitize_html(attribute)
```
jeremyf: With line 42, lines 43 to line 45 are not necessary.

jeremyf:
Excellent stuff! I'm glad that you had a chance to dig through sanitization of user input. Its something we have to be ever vigilant about because our applications more and more hold the keys to so much information.

PR #525, opened 2015-04-22
523 create a work area show page

### LINKED ISSUE #523: Create a work area show page
Given that this is delving into the controller, I'd like to explore a refinement of the runner interaction with the controller; That is to say the callbacks are helpful but have a lot of similarities that can be encapsulated inside a class.

jeremyf:
## Claiming issue 523

@94495352c3159b344aeea60cb4bf71fdb35fcb09

relates to #523

$ ./script/start-issue 523

[skip ci]

## Adding HandledResponseParameter

@a8ce71bcf23b297fe02357973b0f5864cc866456

I want to codify the expectation between a runner's returned value and
a response handler.

## Adding ResponseHandlers module

@3e8dbfac2cb72e08331f5e054c934ace3159f048

> ResponseHandlers are a means of encapsulating how we respond to an
> action's response. Action responses should implement the interface
> of the HandledResponseParameter.
>
> This is experimental; After conversations with @terrellt I wanted to
> eliminate the callbacks of the runners. They were helpful to convey
> what was happening. I believe the callbacks have reduced some of the
> complexities of controller testing. There was room for more, and I
> hope that is made clear through this experimentation.

## Adding exception for interface expectation error

@63210026cba5646d0676b4d388a62ae7f3cd4c75

In cases where we expect something to be implemented but it is not,
this is a case that we want to raise an exception. It highlights a
broken development assumption.

## Adding Work Area response handlers

@8cdf378cbcc8a5facf5a4eb1d66bd68692c49ea0

This is an experiment of treating the controller as the inflection
point of the request and the response but making both the request
logic (i.e. the Runners) and the response logic (i.e. the
ResponseHandler) leverage indirection to better allow for
configuration and testability.

## Creating the WorkArea experiment

@f20bb7c82473ac573bb41e3742f9a7c8f6c49c68

Instead of using the current convention of a Runner and inline
response handling, I'm opting to create a custom respond handling eco-
system.

This should be considered experimental, but I'm enjoying the greatly
simplified controller testing. I may need to inject a dummy responder
that doesn't care about view rendering and such non-sense.

## Closing issue 523

@437b691c21841844040457bf265a3376bd0d209d

Closes #523

$ ./script/close-issue 523


rbalekai:
:+1:

PR #545, opened 2015-04-23
Crafting application concept

jeremyf:
## Updating work_areas routing as per conversation

@fb0bccf049c727bbe4c16817edf66b52d9cd1c2e

Keeping it simpler

## Adding Sipity::Models::ApplicationConcept

@d23e23b47f6f86cde634f611af5704de829d4ee1

> Experimental
>
> Throughout the application I have the case in which I need to relate
> data-concepts to class concepts; (i.e. WorkArea's have the same
> processing strategy).
>
> As I'm thinking about them, I believe the WorkType and the
> ApplicationConcept are in close alignment. It just happens that a
> WorkType already exists and does not make sense to shoe horn the
> WorkArea into the WorkType concept.
>
> Perhaps as I work through this concept, I'll emerge with a better
> abstract representation of both WorkTypes and the WorkArea concept.

## Experimenting with WorkArea ApplicationConcept

@a699b615dd708f169f7e6db1e62512db49c99a23

I am explicitly deciding that a WorkArea application concept is where
I hang the application processing strategy instead of an individual
WorkArea.

The idea being if each WorkArea had different strategies, I believe
they would all be copies of each other. In reality, I want them to
share the strategy and let the implementation forms dictate what each
strategy does.

## Amending method name

@b64f5314278f982f446e220fd2e66711a7c3d74b

As per @rbalekai's suggestion, I want it to be more descriptive.

## Permitting 'show' of new work area to its manager

@981c775f014202ca3d1da1df7956d80e3ea91bcd

Prior to this commit, the work area was in accessible to everyone. Now
defaulting the behavior to grant permission for the work area manager
to "show" the given work area.

## Adding a more exhaustive test case

@ce5025e18a4047e588613844861cf7d5f0b7f492

With the WorkAreas all having the same processing strategy, the
creation behavior needed to account for that.

## Removing :through relationships

@29b202631f4795d3439f77f077319802efc89213

While these may be helpful from a data structure stand point, I
believe these relations are less useful given the preference of
Repository objects.

Also, it clears up the lines that are generated in the ERD.

## Adding declared relationship

@7b711ae596461f61cccb945a367b55b9582b8be1

While it may not be necessary (and perhaps in violation of module
boundaries) to have the has_one declaration for a polymorphic
relation, I want to expose this to assist in the ERD generation.

## Updating Sipity ERDs to reflect changes

@dd3169dd245ad21f6ea9309ee42756da53b9f486


app/services/sipity/services/create_work_area_service.rb
```diff
@@ -33,18 +34,28 @@ def call
       private
 
       attr_accessor :work_area
-      attr_reader :processing_strategy, :work_area_managers
+      attr_reader :processing_strategy, :work_area_managers, :application_concept
 
       delegate :slug, to: :work_area
 
+      def find_or_create_the_work_area_concept!
```
rbalekai: Shouldn't the method name be find_or_create_the_application_concept! as you are calling the Models::ApplicationConcept.create! ?

jeremyf: Good catch. Will change.

On Thursday, April 23, 2015, rbalekai <notifications@github.com> wrote:

> In app/services/sipity/services/create_work_area_service.rb
> <https://github.com/ndlib/sipity/pull/545#discussion_r29001258>:
>
> >
> >        delegate :slug, to: :work_area
> >
> > +      def find_or_create_the_work_area_concept!
>
> Shouldn't the method name be find_or_create_the_application_concept! as
> you are calling the Models::ApplicationConcept.create! ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ndlib/sipity/pull/545/files#r29001258>.
>


-- 

Jeremy Friesen
takeonrules.com

PR #579, opened 2015-04-28
562 Step 2 create processing strategy usage class

### LINKED ISSUE #562: Create Processing::StrategyUsage class
Instead of a ProcessingStrategy being owned by a concept, it should be associated with a concept. In that way we can more readily reuse strategies (even copying them if need be). I believe this will help with fishing around with the ApplicationConcept.

jeremyf:
## Removing Processing::Strategy.proxy_for relationship

@6a531f3b96c9f720ddb1e4e205f1ab918d559355

Tests are broken but this is the moment in which they all break. Now
to rebuild from there.

## Refactoring to remove belongs_to

@4a037d903515b9b9b032f6c005e42f30d5a93014

## Fixing seed data for build

@0e81f2302d0337046f41bc6d7b7fe8640cf91847

## Removing deprecated method test

@bd2dd2709485f0eeab6be8fb1a15abd5f9701c45

## Refactoring to use CreateWorkService

@a28035b4cee3d7c47c5ff7bf6b916d4de1453ca7

Instead of the complicated repository method, I'm favoring a service
object that can better compose existing behavior. The tests continue
to fail as this is simply an extract class from method.

This begs the question:

> Why do some repository methods delegate to a service object and
> other methods have inline behavior?

My short answer is that methods are very readable, but classes allow
encapsulation of ideas. So, as a repository method gets more
complicated it becomes a primary candidate for factoring into a class.

Another way to think about it is that repository methods provide a
convenience method for Presenter and Form interaction with the data.

My suspicion is that each form should leverage a collaborating service
class instead of service method(s); The service class could be swapped
out as well. But forms are a complicated critter; They need data from
the persistence layer and need to issue a command to update the
persistence layer. (They leverage both commands and queries).

With the separation of CommandRepository and QueryRepository, our code
is at a point where forms could be composed of a QueryService and
CommandService object. And that is how things may move going forward.
But for now we factor towards an understanding of how our code is
growing and taking shape.

## Adding FindOrCreateWorkTypeService

@b7346f56f5318c088275496e6a8625f245f8c32a

> Responsible for finding or creating the corresponding work type and the
> requisite processing entries.

Also consider that the existing tests are using a piecemeal approach
to work type creation (and strategy creation). This is a class that
can be leveraged instead of a FactoryGirl construct (which would not
be leveraged in production).

## Factoring what happens for Work creation

@8f81ad20cd822678493d31d6da6d7e2898349246

Leveraging composition over imbedded complex behaviors

## Fixing create processing strategy

@7ab7709a0056d5ba30efde11f12c2ecd2d6f5466

This was broken after I removed the proxy_for_id and proxy_for_type
columns from the 'sipity_processing_strategies' table.

## Removing deprecated methods

@d714e5885da714d32d471c6f85f293e282a6eb92

They were no longer being called, so out the door they go!

## Preferring service vs. direct model create! calls

@256bac13e901ee3a54d6b6feafb05e5ce740be21

There is a service class that collaborates on the data creation for a
work type and such. I want to leverage it instead of "direct"
statements for creating the various database entries.

## Closing issue 562

@73596810b6a69e01a0cad5f72ae92ca2f3c237b2

Closes #562

$ ./script/close-issue 562


rbalekai:
the tests ran successfully on my dev for this PR

Issue #584, opened 2015-04-29
Create an "All Registered Users" Group and associate all users to it on create


jeremyf:
- [ ] Create a ALL_REGISTERED_USERS constant in the Sipity::Models::Group
- [ ] Seed an "All Registered Users" group 
https://github.com/ndlib/sipity/blob/master/db/seeds/groups_and_roles_seeds.rb
- [ ] In the same seed, add all existing users to that group (unless they are already there); (Consider a custom query for this)
- [ ] Create a callback for user on create that will add them to the ALL_REGISTERED_USERS; Instead of a single method, create a callback class (`DataGenerators::OnUserCreate`) that encapsulates all of the data entries that should happen.

I'm thinking that the callback class would act like a function; That is to say in User, you would do `DataGenerators::OnUserCreate.call(user: self)`. I also believe that we would want to use a [class_attribute](http://api.rubyonrails.org/classes/Class.html#method-i-class_attribute) to associate the constant below is a possible implementation guide-post:

```ruby
class User
  self.class_attribute :on_user_create_service
  self.on_user_create_service = Rails.application.config.default_on_user_create_service
  
  after_commit :call_on_create_user_service, on: :create
  private
  def call_on_create_user_service
    on_user_create_service.call(self)
  end
end

PR #586, opened 2015-04-29
584 create an all registered users group and associate all users to it on create

banurekha:
Could you please review the changes and comment? Everything as part of this issue #584 should be covered. Let me know if I miss something


app/data_generators/sipity/data_generators/on_user_create.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+
+      attr_reader :user
+
+      private
+
+      attr_writer :user
+
+      def add_user_to_registered_group!
+        register_user = Sipity::Models::Group.find_or_create_by!(name: Models::Group::ALL_REGISTERED_USERS)
```
jeremyf: The local variable is perhaps misnamed; When I was reading it I thought you were referring to a user. Is there a clearer variable name? `all_registered_users_group` perhaps?

banurekha: yes! it is misnamed.

config/application.rb
```diff
@@ -31,6 +31,10 @@ class Application < Rails::Application
       Services::NoidMinter.call
     end
 
+    config.default_on_user_create_service = lambda do
```
jeremyf: In reviewing how you are using this lambda, the class won't work. I'm going to expand the code to clarify:

```ruby
class User
  self.on_user_create_service = Rails.application.config.default_on_user_create_service
  # Is the same as
  self.on_user_create_service = lambda { DataGenerators::OnUserCreate }
  # Is the same as
  def call_on_create_user_service
    lambda { DataGenerators::OnUserCreate }.call(self)
  end
end
```

Which means that the configuration as written will never call the `DataGenerators::OnUserCreate` class method.

spec/data_generators/sipity/data_generators/on_user_create_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+        it 'will instantiate then call the instance' do
+          expect(described_class).to receive(:new).and_return(double(call: true))
+          described_class.call(user: user)
+        end
+      end
+
+      it 'will add user to register)user_group' do
+        expect { described_class.call(user: user) }.
+          to change { Models::Group.count }.by(1)
+      end
```
jeremyf: There is no verification that the user has been added to the newly created group.

config/application.rb
banurekha: @jeremyf so I have to modify application.rb as
 config.default_on_user_create_service =  DataGenerators::OnUserCreate

jeremyf: I believe that is what needs to be done; Though you'll need to make it `Sipity::DataGenerators::OnUserCreate`

banurekha: After doing some querying and using console, only thing that seems to work is 
config.default_on_user_create_service = lambda do |user|
      Sipity::DataGenerators::OnUserCreate.call(user)
    end
Anything else gives error on start up or argument error

db/seeds/groups_and_roles_seeds.rb
```diff
@@ -4,6 +4,13 @@
   Sipity::Models::Group.find_or_create_by!(name: GRADUATE_SCHOOL_REVIEWERS)
 )
 
+$stdout.puts 'Add existing users to All Registered Users Groups...'
+Sipity::Models::Group.find_or_create_by!(name: Sipity::Models::Group::ALL_REGISTERED_USERS)
+#add existing users to that group (unless they are already there)
+Sipity::Models::User.all.each do |user|
```
jeremyf: Prefer `Sipity::Models::User.find_each` instead of the more memory intensive `Sipity::Models::User.all.each`

spec/data_generators/sipity/data_generators/on_user_create_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+          described_class.call(user: user)
+        end
+      end
+
+      context 'add new user to all_registered user group' do
+        let(:user) { Sipity::Factories.create_user }
+        let(:all_registered_users_group) { Sipity::Models::Group.find_by(name: Models::Group::ALL_REGISTERED_USERS) }
+        it 'user will be part of all_registered user group' do
+          described_class.call(user)
+          expect(all_registered_users_group.group_memberships.where(user: user).map(&:group)).
```
jeremyf: The reason is that you are retrieving everything, then iterating on everything, and loading each of those groups; That's 3 loops with 2 more expensive queries vs. the counts single query.

app/data_generators/sipity/data_generators/on_user_create.rb
```diff
@@ -0,0 +1,30 @@
+module Sipity
+  module DataGenerators
+    # Responsible for adding user to registered user group
+    #
+    class OnUserCreate
+      def self.call(user)
+        new(user: user).call
+      end
+
+      def initialize(user:)
```
jeremyf: I'd prefer to see the `self.call` method signature look more like the `initialize` method signature. The external behavior is to have an unnamed parameter. The reason for the preference is that it shows that the two methods have a common enough interface and one that doesn't change; So I can initialize and call the class with the same parameters.

So go with:

```ruby
def self.call(user)
  new(user).call
end

def initialize(user)
  self.user = user
end
```

PR #595, opened 2015-04-30
588 create work form needs to be made specific to etd

### LINKED ISSUE #588: CreateWorkForm needs to be made specific to ETD
With the close of #573 we have a "place" to associate the valid work types for ETDs.

Once #587 is merged, my short-term proposal is to [change the following](https://github.com/ndlib/sipity/blob/fa8a866f2ec0f6abfbeef3b9247d6f6087f74cb2/app/forms/sipity/forms/create_work_form.rb#L69-L71):

```ruby
def possible_work_types
  Models::Work.work_types
end
```

To leverage `Sipity::DataGenerators::Etd::WorkTypesProcessingGenerator::WORK_TYPE_NAMES`; This is an interim solution, but ensures that as we add the `ULRA Submission` work type that we do not end up with an accidental crossing of `WorkArea` boundaries.

banurekha:
Closes #588


app/forms/sipity/forms/create_work_form.rb
```diff
@@ -35,7 +35,7 @@ def work_publication_strategies_for_select
       end
 
       def work_types_for_select
-        possible_work_types.map { |elem| elem.first.to_sym }
+        possible_work_types.map(&:to_sym)
```
jeremyf: Can you see about replacing `:to_sym` with `:to_s`. Symbols are never collected via Garbage Collection in the system (not quite true with Ruby 2.2.x but something to be mindful of).

app/models/sipity/models/work.rb
```diff
@@ -41,8 +41,7 @@ def to_s
           ALREADY_PUBLISHED => ALREADY_PUBLISHED,
           GOING_TO_PUBLISH => GOING_TO_PUBLISH,
           DO_NOT_KNOW => DO_NOT_KNOW
-        },
-        work_type: WorkType.all_for_enum_configuration
+        }
```
jeremyf: Why did you remove this? I believe this is something that needs to be enforced.

banurekha: @jeremyf only place it was used was https://github.com/ndlib/sipity/blob/master/app/forms/sipity/forms/create_work_form.rb#L70 and we are refactoring that. So I decided to remove work_type enum. Should I bring it back? not sure what it is enforcing

app/forms/sipity/forms/create_work_form.rb
banurekha: We use Simple_form for 18n translation for work_type values in collection and simple_form expect them to be sym rather than string. I guess in this case it has to be to_sym

jeremyf: That jogs my memory, could you add a comment on the `.to_sym` behavior?

app/models/sipity/models/work.rb
jeremyf: Without the enum, valid work_types are no longer a controlled vocabulary (meaning someone could create a `GrizzlyBear` work type by doing DOM injection. I don't know what would then happen, but I'd prefer an exception as this would be something exceptional.

spec/models/sipity/models/work_spec.rb
```diff
@@ -27,27 +27,6 @@ module Models
         end
       end
 
-      context '.work_types' do
```
jeremyf: Please restore these specs.

Issue #617, opened 2015-05-06
Remove default submission window from create a work service


jeremyf:
Based on the work related to #616, I want to associate the creation of a work with a work area and submission window.

The conceit of #616 is that there is a default submission window (ETD's Start). That doesn't hold water as we proceed.

PR #618, opened 2015-05-06
616 associate a work and the submission window under which it was created

jeremyf:
## Adding WorkArea specific form for create a work

@4fade2c42b35fa9d93e8190664cb9aef191dfbe8

I want to begin associating works with submission windows. This is the
first step saying that the existing form building should go through
the assumed ETD system.

There will need to be a builder going forward that makes a
determination on all forms to find. The existing setup is somewhat
ad-hoc but can be made more sane.

## Removing a feature test

@f2e0c3482d6f3a517c759bcb4dbdf3cc90beff9a

This feature test, while important, is handled via numerous other unit
tests and the layers of interaction.

It is also something that is better managed outside of the code-base
as it reflects configuration options that the test suite is perhaps
somewhat agnostic towards.

## Catching a missing spec

@a8ffb4195d49392f3d4e37804dc7c8c3c1dcbaa9

With the removal of a feature test (@f2e0c34), it appears as though
the test coverage has dropped.

Adding a unit test to make sure things are working reasonably well.

## Adding WorkSubmission

@7b186cc77073ba8f366c64dd4cfb2366d390d40d

Instead of opting for a Work belongs to a SubmissionWindow, I'm
instead preferring that all works are associated with a
WorkSubmission. In doing so, the idea is that a Work can exist outside
of a Submission Window; That is to say if we delete the Submission
Window that the Work was submitted into, we don't want to delete the
Work.

## Creating a work associates with submission window

@49b0219f53021b36621ecddd2ab2677482afd81d

At present the :submission_window is an optional parameter. If none is
provided things default to the etd/start
`work_area/submission_window`.

At a future point, as we splice in the ULRA forms, this default will
be removed. See #617 for reference to that work.


banurekha:
:+1:

PR #619, opened 2015-05-06
Removing AccountPlaceholder from Sipity

jeremyf:
There are a lot of assumptions about the AccountPlaceholder behavior
that may or may not hold true. At present it is not being used; And I
believe the use case may not directly map to what was implemented.

If we find that we want to proceed with DOIs there are many
salvageable bits of code. However, I don't want to have to keep
carrying along code that is not leveraged by production.


banurekha:
:+1:

PR #620, opened 2015-05-06
Moving a form into Forms::Core

jeremyf:
> A container for classes that are core to Sipity; That is to say
> things I don't believe to be part of a potential plugin.
>
> Why not Base? I believe Base to be more representative of things you
> build from, not stand alone elements.


banurekha:
:+1:

PR #664, opened 2015-06-02
Amending uncalled method check

jeremyf:
## Checking config for uncalled methods

@de74f97cb2f69df48a962766a8dd22e1cb8d996e

Lots of things happen in config, lets see how things are going.

[skip ci]

## Amending uncalled list

@08c923166544274b23531cf8065a4d6c8820f863

Leveraging changes from the following pull requests:

* seattlerb/debride#19
* seattlerb/debride#18

*I had to build a local version of the debride gem with these
changes.*

[skip ci]

### DIFF
--- a/scripts/uncalled-method-check
+++ b/scripts/uncalled-method-check
@@ -58,7 +58,7 @@ if `cd #{REPOSITORY_PATH} && git status --porcelain`.strip.size > 0
 end
 
 begin
-  command = "cd #{REPOSITORY_PATH} && debride -r lib app --whitelist #{WHITELIST_PATH} > #{OUTPUT_FILENAME}.tmp"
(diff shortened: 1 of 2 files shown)


scripts/uncalled-method-check
```diff
@@ -58,7 +58,7 @@ if `cd #{REPOSITORY_PATH} && git status --porcelain`.strip.size > 0
 end
 
 begin
-  command = "cd #{REPOSITORY_PATH} && debride -r lib app --whitelist #{WHITELIST_PATH} > #{OUTPUT_FILENAME}.tmp"
+  command = "cd #{REPOSITORY_PATH} && debride -r lib app config --whitelist #{WHITELIST_PATH} > #{OUTPUT_FILENAME}.tmp"
```
zenspider: I don't recommend running debride on `config/routes.rb` as that can lead to false positives. I've recently added (unreleased so far) `bin/debride_rails_whitelist` which lets you empirically generate a whitelist from production logs. This should be released soon and allow rails sites to generate whitelists based on actual data.

PR #787, opened 2015-06-22
603 map work attributes and its related attachments to curate etd and generic file content model 

banurekha:


app/exporters/sipity/exporters/etd_exporter.rb
```diff
@@ -0,0 +1,50 @@
...
+        self.work = work
+        self.repository = repository
+        self.attachments = repository.work_attachments(work: work)
+      end
+
+      def call
+        # Create rof etd file to be ingested
+        file_name = ROF_FILE_PREFIX + work.id + ROF_FILE_EXTN
+        FileUtils.mkdir_p(MNT_PATH) unless File.directory?(MNT_PATH)
+        rof_file = File.new(File.join(MNT_PATH, file_name), 'w')
```
jeremyf: Use the following as it is better with handling exceptions and ensuring the file is flushed and closed; The implementation that you are using will keep the file open if an exception is thrown during the process.

```ruby
File.open(File.join(MNT_PATH, file_name), 'w+') do |rof_file|
  rof_file.puts '[' + export_to_json.join(',') + ']'
end
```

app/mappers/sipity/mappers/etd_mapper.rb
```diff
@@ -0,0 +1,180 @@
...
+      def initialize(work, repository: default_repository)
+        self.work = work
+        self.repository = repository
+      end
+
+      def call
+        build_json
+      end
+
+      def build_json
```
jeremyf: Why the indirection of `call` then calling `build_json`; It looks as though `build_json` should be a private method.

app/mappers/sipity/mappers/etd_mapper.rb
```diff
@@ -0,0 +1,180 @@
...
+          end
+          json.set!(PROPERTIES_KEY, "<fields><depositor>#{BATCH_USER}</depositor></fields>")
+        end
+      end
+
+      private
+
+      attr_accessor :work, :repository
+
+      def append_namespace_to_id
```
jeremyf: The method name `append_namespace_to_id` feels as though it will be changing the `work.id`. The implementation of the method strikes me as though the method name should be something like `namespaced_pid`.

app/mappers/sipity/mappers/etd_mapper.rb
```diff
@@ -0,0 +1,180 @@
...
+        NAMESPACE + work.id
+      end
+
+      def attributes
+        @attributes = {}
```
jeremyf: Why is this an instance_variable? Its not memoized, but instead rebuilt each time the method is called.

app/mappers/sipity/mappers/etd_mapper.rb
```diff
@@ -0,0 +1,180 @@
...
+      end
+
+      def embargo_date
+        embargo_dates = work.access_rights.map(&:transition_date)
+        embargo_dates.map { |dt| dt.strftime('%Y-%m-%d') }
+        embargo_dates.to_sentence
+      end
+
+      def extract_name_for(attribute)
+        ETD_ATTRIBUTES[attribute]
```
jeremyf: Why does this method use `ETD_ATTRIBUTES[attribute]` and the other uses `ETD_ATTRIBUTES[attribute.to_s]`?

config/initializers/etd_mapper_initializer.rb
```diff
@@ -0,0 +1 @@
+ETD_ATTRIBUTES = YAML.load( File.open( Rails.root.join( 'config/etd_mapper.yml' ) ) ).freeze
```
jeremyf: In adding these as an initializer, we are slowing down our boot time. Can you create a Hash constant inside the Exporters namespace?

I want to get away from the proliferation of Yaml files.

spec/mappers/sipity/mappers/etd_mapper_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+        expect(repository).to receive(:work_attribute_values_for).with(work: work, key: 'defense_date').and_return([])
+        expect(repository).to receive(:work_attribute_values_for).with(work: work, key: 'program_name').and_return(['a program_name'])
+        expect(repository).to receive(:work_access_right_codes).with(work: work).and_return(access_right)
+        expect(repository).to receive(:scope_users_for_entity_and_roles).
+          with(entity: work, roles: 'creating_user').and_return(creators)
+        expect(work).to receive(:id).and_return('a_id')
+        expect(work).to receive(:title).and_return(title)
+        expect(work).to receive(:collaborators).and_return(collaborators)
+        expected_json = JSON.parse(subject.call)
+        puts "Json: #{expected_json.inspect}"
```
jeremyf: Remove this `puts` statement.

app/mappers/sipity/mappers/generic_file_mapper.rb
```diff
@@ -0,0 +1,208 @@
...
+      end
+
+      def file_name_to_create
+        return file.pid + '-' + file.file_name
+      end
+
+      def decode_access_rights
+        # determine and add Public, Private, Embargo and ND only rights
+        decoded_access_rights = { READ_KEY => creators,  EDIT_KEY => [BATCH_USER] }
+        access_rights.each do |access_right|
```
jeremyf: What happens if there are two access rights in the array? It looks as though things will get obliterated. Should we be assuming that there will be only one access rights entry per attachment / work?

app/mappers/sipity/mappers/generic_file_mapper.rb
```diff
@@ -0,0 +1,208 @@
...
+      end
+
+      def embargo_date
+        if file_access_rights.empty?
+          embargo_dates = work.access_rights.map(&:transition_date)
+        else
+          embargo_dates = file.access_rights.map(&:transition_date)
+        end
+        embargo_dates.map { |dt| dt.strftime('%Y-%m-%d') }
+        embargo_dates.to_sentence
```
jeremyf: I understand that we have an array, but does it make sense that we would have multiple dates asserted as a sentence in our Hydra Rights Metadata?

app/models/sipity/models/attachment.rb
```diff
@@ -30,6 +30,7 @@ def thumbnail_url(width: THUMBNAIL_WIDTH, height: THUMBNAIL_HEIGHT)
       end
 
       delegate :url, to: :file, prefix: :file, allow_nil: true
+      delegate :path, to: :file, prefix: :file, allow_nil: true
```
jeremyf: You can add `:path` to line 32, as follows:

```ruby
delegate :url, :path, to: :file, prefix: :file, allow_nil: true
```

config/application.yml
```diff
@@ -30,8 +30,14 @@ development:
...
   secret_key_base: 'bd6953190e19dabc215af75a5c856f584dca49a3ab729696946a80ec50f46094d4643d908b88b07011f898f5c6748f2ffa832c96bae409e5462f2cf3055f1ee7'
+  curate_batch_user_pid: "batch_user_pid"
+  curate_batch_group_pid: "batch_group_oud"
+  curate_batch_mnt_path: "local/path/to/create/rof/file"
 
 test:
   domain_name: "localhost:3000"
   url_host: "http://localhost:3000"
   secret_key_base: 'bd6953190e19dabc215af75a5c856f584dca49a3ab729696946a80ec50f46094d4643d908b88b07011f898f5c6748f2ffa832c96bae409e5462f2cf3055f1ee7'
+  curate_batch_user_pid: "batch_user_pid"
```
jeremyf: Are these three keys required via Figaro? If so add them as required to `config/initializers/figaro.rb`

app/mappers/sipity/mappers/generic_file_mapper.rb
banurekha: I think multiple embargo dates are not supported in Hydra Rights metadata, so I would like to grab first but not sure what is the right way to do it.. I had the same question when I wrote this..

app/mappers/sipity/mappers/generic_file_mapper.rb
banurekha: Thats is my assumption.. At this point it looks there can be only one access rights per work/attachment.. Not sure if we need to support access rights array.. Open for suggestions..

app/mappers/sipity/mappers/generic_file_mapper.rb
jeremyf: I'd feel better if you assumed `.first`; That way we don't have a mangled embargo date on our server.

Maybe the access_rights relationship should change from `has_many` to `has_one`

PR #794, opened 2015-06-25
Correction in ETD vocabulary control

rbalekai:


jeremyf:
@rbalekai You may need to update locabulary as well. https://github.com/ndlib/locabulary

PR #830, opened 2015-07-07
828 can not delete collaborator

jeremyf:
Cleaning up DOM when "Remove" repeating field

The following image demonstrates the DOM state before the "Remove"
button is pressed. The input[type="hidden"] property

repository_deposit_interface

Please note that this is a kludge type solution, and most certainly
not a long term solution. The comments in code includes possible solutions.

> In the case of fields_for, a hidden input field is added that allows for
> passing the object identifier as part of the form submission. This also
> needs to be removed when we remove the repeating field.
>
> This is a kludge type hack in that what is really needed is to look at the
> input fields of the parent, find its DOM-ID prefix (i.e. For work_0_name, the
> prefix would be work_0). Then find the corresponding inputs with that same prefix
> and _id (i.e. work_0_id); We would then delete those elements.


danhorst:
:+1:

PR #845, opened 2015-07-10
820 email language changes for confirmation of grad school signoff

rbalekai:


app/decorators/sipity/decorators/emails/work_email_decorator.rb
```diff
@@ -42,6 +58,30 @@ def email_message_action_url
           view_context.work_url(work)
         end
 
+        def program_names
```
jeremyf: I would prefer that these methods be accessed via a repository method. You have a subtle violation of the Law of Demeter:

```ruby
work.additional_attributes.program_name.map(&:value)
```

```ruby
repository.work_attribute_values_for(work: work, value: Models::AdditionalAttribute::PROGRAM_NAME)
```

Also, consider do you want these to be arrays?

jeremyf:
For reference, you can use  `$ rake spec:coverage:decorators` to see how the decorators are covered in isolation.

PR #856, opened 2015-07-22
850 display the repository url during cataloging

banurekha:


app/presenters/sipity/controllers/work_submissions/show_presenter.rb
```diff
@@ -94,6 +94,14 @@ def section(identifier)
           TranslationAssistant.call(scope: :sections, subject: work_submission, object: identifier)
         end
 
+        def repository_url_label
+          TranslationAssistant.call(scope: :sections, subject: work_submission, object: 'repository_url')
+        end
+
+        def repository_url_for_work
+          Figaro.env.curate_nd_url! + work_submission.id
```
jeremyf: Could you use `File.join(Figaro.env.curate_nd_url!, work_submission.id)`? It ensures that we have a "/" between the two values.

PR #857, opened 2015-07-22
855 add java script to prevent double click on submit button

banurekha:


app/assets/javascripts/application.js
```diff
@@ -40,11 +40,17 @@
     }
   };
 
+	var disableSubmitOnClick = function(){
+		var submitButton = $(":input[type='submit']");
+		submitButton.attr('data-disable-with', 'Please wait...');
```
jeremyf: Should this be for all forms that have an action that is not GET?

PR #890, opened 2015-10-01
Make scheduled_time field optional

rbalekai:
This was requested by Grad School on 09-29-15

### DIFF
--- a/app/forms/sipity/forms/work_submissions/etd/ingest_with_postponed_cataloging_form.rb
+++ b/app/forms/sipity/forms/work_submissions/etd/ingest_with_postponed_cataloging_form.rb
@@ -21,7 +21,6 @@ module Sipity
 
           include ActiveModel::Validations
           validates :agree_to_signoff, acceptance: { accept: true }
-          validates :scheduled_time, presence: true
 
           attr_reader :scheduled_time, :reason
 
@@ -35,7 +34,7 @@ module Sipity
           end
 
           def add_scheduled_time(f:)
-            f.input(:scheduled_time, input_html: { required: 'required' }).html_safe
+            f.input(:scheduled_time, input_html: { value: work.access_right_transition_date }).html_safe
           end
 
           def add_agree_to_signoff(f:)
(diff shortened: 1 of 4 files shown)


app/forms/sipity/forms/work_submissions/etd/ingest_with_postponed_cataloging_form.rb
```diff
@@ -35,7 +34,7 @@ def render(f:)
           end
 
           def add_scheduled_time(f:)
-            f.input(:scheduled_time, input_html: { required: 'required' }).html_safe
+            f.input(:scheduled_time, input_html: { value: work.access_right.transition_date }).html_safe
```
jeremyf: I don't like a violation of the Law of Demeter in the view rendering logic.

Can you add to Sipity::Models::Work the following:

```ruby
  delegate :transition_date, to: :access_right, allow_nil: true, prefix: true
```

And the corresponding spec:

```ruby
it { should delegate_method(:transition_date).to(:access_right).prefix(:access_right) }
```

Then change `work.access_right.transition_date` to `work.access_right_transition_date`.

spec/forms/sipity/forms/work_submissions/etd/ingest_with_postponed_cataloging_form_spec.rb
```diff
@@ -67,6 +72,28 @@ module Etd
...
             end
           end
+
+          context 'with out scheduled_time' do
+            subject do
+              described_class.new(
+                keywords.merge(
+                  attributes: {
+                    agree_to_signoff: true,
+                    scheduled_time: nil
```
jeremyf: More likely that you will get an empty string `""` instead of `nil`. Test that.

The reason for the empty string is that an HTML form's input will submit empty strings if no values exist.

app/forms/sipity/forms/work_submissions/etd/ingest_with_postponed_cataloging_form.rb
```diff
@@ -61,12 +60,15 @@ def signoff_agreement
           end
 
           def submit
-            processing_action_form.submit do
-              repository.create_scheduled_action(
-                work: work,
-                scheduled_time: scheduled_time,
-                reason: Sipity::Models::Processing::AdministrativeScheduledAction::NOTIFY_CATALOGING_REASON
-              )
+            if scheduled_time.present?
```
jeremyf: Make a method `create_scheduled_action_if_applicable`

Then change the submit method to:

```ruby
def submit
  processing_action_form.submit do
    create_scheduled_action_if_applicable
  end
end

def create_scheduled_action_if_applicable
  return true unless scheduled_time.present?
  repository.create_scheduled_action(
    work: work, scheduled_time: scheduled_time, reason: Sipity::Models::Processing::AdministrativeScheduledAction::NOTIFY_CATALOGING_REASON
  )
end
```

PR #1019, opened 2016-02-09
Fixing catalog system number coercion

jeremyf:
A weird scenario in which the value "004316606" was converted to
"001154438". Not sure why, but I'm adding a guard condition for that
scenario.


dbrower:
I suspect the original number "004316606" was parsed as octal. (that number in octal is equal to the number 1154438 in decimal). Maybe the input parser accepts a leading 0 as indicating octal?

jeremyf:
@dbrower That was my assumption and the consensus of Twitter.

PR #1037, opened 2016-03-03
Extracting power converters to separate files

jeremyf:
## Extracting individual power converters to files

@edd55149a8345609310082607465fa406cd482c7

Instead of the Frankenstein's Monster's big list of conversions, focus
on creating a directory that has each of them.

## Isolating all power converter specs

@ee1091f2d7ea9d339bbbb764e89f7beab1cc63e0

Instead of relying on a massive test suite allow each conversion to be
tested in isolation.


jeremyf:
Yeah. I'm doing some preparation for work that I want to do with Larita.

PR #1055, opened 2016-03-30
Adding minor revisions to BatchIngestExporter

jeremyf:
* Favor using the attr_writer method instead of using an instance
  variable. By leveraging a setter method, we can add an interface
  check when we attempt to set the work. We also don't rely on internal
  state of the object but instead prefer to consume the internal API of
  that object.
* Privatize access to the `work` object. It does not appear that `work`
  is referred to elsewhere. Keeping the exposed methods to a minimum
  allows for greater flexibility later on. (i.e. you are not committing
  to any more of the public interface than required)
* Replace `work.id` with `#work_id`. This is a move for greater
  consistency in method usage.

Question does the privatization mean we should be passing the
exporter's work as a parameter? Or should we leave it exposed?

### DIFF
--- a/app/exporters/sipity/exporters/batch_ingest_exporter.rb
+++ b/app/exporters/sipity/exporters/batch_ingest_exporter.rb
@@ -11,10 +11,14 @@ module Sipity
       end
 
       def initialize(work:)
-        @work = work
+        self.work = work
       end
 
+      private
+
       attr_accessor :work
+
+      public
 
       def call
         AttachmentWriter.call(exporter: self)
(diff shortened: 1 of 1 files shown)


danhorst:
I wasn’t sure where the best place for `#make_data_directory` would be. I made it a part of `BatchIngestExporter` to avoid coupling `WebhookWriter.call` to another subclass of `BatchIngestExporter` that also needs the method, like `AttachmentWriter`.

I’m not sure I like passing both the exporter _and_ the work as paramters. I was thinking of the exporter as a wrapper around a single work. I suppose this is naïeve as it preculdes its use on groups of works.

danhorst:
I like the API refinements you proposed.

PR #1078, opened 2016-04-21
ETD need json end point for all dissertations that are submitted for review through ingested

### LINKED ISSUE #1062: ETD - Need JSON end point for all dissertations that are submitted for review through ingested
Based on conversations with the Grad School and the OIT we need to provide a single HTTPS end point that:

1) Is accessible to the ETD Reviewer (i.e. Shari) or a hand crafted user token for OIT
2) Is scoped to the ETD work area (`/areas/etd/do/list_submissions.json`)
3) Returns an Array of JSON objects that are in a to be defined "ETDs for the OnBase document integration" format

If designed based on existing permission queries, we can open this JSON end point to all users.

See [Data Map document](https://docs.google.com/spreadsheets/d/1CPCe5Y_vnnpUlpU2PXdwimHdAoD-sYxgCvbHAxdBbIQ/edit#gid=0)

jeremyf:
## Claiming issue 1062

@ebe8c033fefe1e186d51c6e1bdad41f0433ebf0a

relates to #1062

$ ./script/start-issue 1062

[skip ci]

## Adding list_submissions action for ETD Work Area

@d78827c83cd37955bca21851f47497ebc51888d4

It can be very helpful to see all of the submissions in a work area.
As such, this is in place regarding that information.

## Ensuring that the application requires repository

@36385cf530948ced16b10203f6238b9d3e45ebed

I was encountering a case in which the Repository object did not load
the modules. By requiring the class as part of the controller, it
appears my problem was solved.

## Adding ability to handle different formats

@5b88613d04bc8f3deb1380099152b89688384bcd

Instead of the default `:html`, we need to be able to handle other
formats. In particular `:json`

## Adding `#scope_attributes_for`

@50ff2a3cf6a348c297e801029f07c310cbb3cbed

I need to pluck all of the attributes instead of singular attributes.
This way things aren't as slow-moving.

## Preliminary addition of ListSubmissions rendering

@63f3d3a97badd40792b9c63f658dcf58fb2cd99d

This is incomplete work, and something I will continue working on to
bring about the full details.

## Adding additional attributes for file access

@56a1e1fd48070cbe10275349c14ee08efc3d7ea0

In the ListSubmissions, I'm adding additional information concerning
the file:

* What is the access right code associated with the object
* What is the release date associated with the object

## Refactoring method to leverage new method

@1d95165569f93b95105f7ef158acd98a6aa4f102

There was a duplication of logic that I wanted to remove.

## Adding #work_attribute_key_value_pairs_for

@44ab2c16c468f4147eb2d69beffd42cc4da03031

Instead of relying on a scoping object as we interact with the
repository method, I want to return a specific data structure. The
reason is that the unit test can then focus on interacting with an
array instead of mocking/testing against an ActiveRecord relation
object.

## Adding ExpandedWork#additional_attributes

@abbb579358f03680ff3e599eddb7f7c827d785ae

Including the additional attributes that exist for each of the works.

## Removing method

@f11c635bb311b34b19bf37a5bb59247e6b1cf5d5

This method was not tested as part of the unit tests. Its available via
a .find_by! call.

## Adding Group#api_key as persisted attribute

@df38b489545f576108eb6af8d813669f07d02632

I have a use case in which I need to register another group. I want to
move the CallbackController authenticate user logic into the database;
But that means providing a space for API keys.

The easiest mechanism is to use a Group.

## Adding Group::ETD_INTEGRATORS

@f45008166db17d4a6fb583f567845fc44f8538b8

I chose all caps and underscore because I'd prefer to not complicate
the integrators life with spaces and such.

## Narrowing who can see the list of submission

@cc30a5753ab1e08b23f5bc436dd954a162b4e3cf

The reason for the narrow scope is because I don't want to have someone
see so many things. At present, Shari would be able to see EVERYTHING
in the ETD system and I would then need to account for pagination.

## Adding show etds to ETD_INTEGRATORS

@dda5da532bacde56c7d409fa3894643a34c73189

As per the requirements of OnBase integration, adding the ability for
the ETD_INTEGRATORS to see the various works that are in the system.
They don't need to see "new" submissions nor "ingested" submissions.

The general idea is that we will have a handful of ETDs that are
working their way through submission; Hopefully no more than 150. As
such, I'm not yet introducing pagination.

However, it may need to be introduced if the system gets bogged down
creating the complicated JSON object.

## Adding Conversions::ConvertToJsonApiPaginationUrl

@2c6cc73b44c52e3881b9f63dc1ed412f06e96c95

I need a consistent manner for generating JSON Api links

## Adding roles that were added to config

@509369cd967a7d619d3107620954a77f637baad1

I missed these during the configuration setup.

## Adjusting ExpandedWork to look like JSON API fmt

@3955a4d77d9cb681948452b545e2c269ead88c1c

I'm looking for inspiration from the JSON API for modeling the data.
The primary reason is that I want to have a consistent pagination
experience.

## Adding show a work via JSON

@fc70f7d46091006bab7dcda52b9526c8aa067621

Given that we are allowing for works to be viewed as a JSON object in
the list_submissions, I want to make sure that we are exposing the JSON
format for the show page.

## Adding JSON Api pagination to list_submissions

@a0195c2f77d36e5b13d8b71ca908c12c980981ce

I don't want our system crushed by large queries. So I introduced
pagination. Its a bit more work for the end-users, but it saves the
server from overloaded work.

## Registering the application/vnd.api+json alias

@ccef3a87e9e261e00092891c688048937fb568bb

## Sorting application.yml keys

@de8baf73a6e8f45ef6a01267301e563cb45207c0

## Updating key name for access_keys

@e22e60a8ed80e6e9b32ab9e7033aa45337c87c43

I wanted to sort them in a meaningful order:

sipity_access_key_for_batch_ingester

1) sipity - provides scope for the application
2) access_key - provides context for the classification of things
3) for_batch_ingester - the specific context for the key

Note: This also involved updating the secrets repository

## Adding ability to extract basic auth string

@5abc90e3f0e9b24b387f72bb49e7eebc802870fc

I need a consistent way to extract basic authentication information. If
none is defined, I want a very custom exception to be raised.

## Moving API key from hard coded into the database

@93447639e87dfce69d4f891d70e62fa2eed526d4

I need more than one group with an API key. As such, I want to have a
more generalized mechanism for working with that information.

## Refactoring method name and params for clarity

@8b5ea89467d8904474327973edc87fa1ca289bee

Again, I'm aiming to generalize the behavior. This means refactoring
for greater clarity of method meaning.

## Defining the ETD Integrator group's keys

@6b20956786e0589f7ab6f362e99e476efb752062

## Generalizing Basic Authentication for Application

@cbb1359972361d970f3090bdf4689b559d04af57

Instead of the authentication existing only for callbacks, exposing the
basic authentication option as a general option.

## Adding an AuthenticatedController

@476f29cb57d2ef373d4a397b9eb336f1bb191b4f

By moving things into the ApplicationController, I was obliterating the
VisitorController's behavior.

## Privatizing a controller method

@fbb633f819882dfc21b65a06a33f18151697a028

There is no reason for this method to be exposed to the world.

## Closing issue 1062

@ca774f82a8d69f3d6efbd80489d885dfcf75625c

Closes #1062

$ ./script/close-issue 1062


spec/forms/sipity/forms/work_areas/core/list_submissions_form_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  module Forms
+    module WorkAreas
+      module Core
+        RSpec.describe ListSubmissionsForm do
+          let(:work_area) { double(name: 'Hello Name', slug: 'hello-world') }
+          let(:repository) { QueryRepositoryInterface.new }
+          let(:user) { double }
+          let(:attributes) { {} }
+          let(:form) { described_class.new(work_area: work_area, requested_by: user, repository: repository, attributes: attributes) }
+          subject {  form }
```
danhorst: Extra space within curly brackets.

PR #1107, opened 2016-08-10
Standardize character used for apostrophe

laritakr:
CurateND used Master's Thesis, while Sipity used Master’s Thesis. This
discrepancy caused a split in the work type facet. Since the majority of the
items were using Curate's method, the decision was made to modify it in Sipity.

Data migration will still be needed on the 47 items that were using this apostrophe.


dbrower:
Is it strange that there is a test already, and it is using a straight single quote? How was this test passing before?

https://github.com/ndlib/sipity/blob/master/spec/features/sipity/export_etd_to_batch_ingest_spec.rb#L37

laritakr:
@dbrower  I'm not sure, unless the rspec tests don't see the difference... Solr query also picked up both types in my query selection when I tested it there.

jeremyf:
@dbrower I'm short-circuiting I18n usage throughout the tests as it greatly increases run time. As a new task, I believe any "feature" specs should leverage I18n. I will add that as a JIRA ticket.

PR #1108, opened 2016-08-10
Change 'restricted' to 'registered'

laritakr:
The read access group for some ETDs were being set to restricted instead of
registered, causing CurateND to treat the record as private.


spec/conversions/sipity/conversions/to_rof/access_rights_builder_spec.rb
```diff
@@ -36,7 +36,7 @@ module Sipity
       end
       context "for Models::AccessRight::RESTRICTED_ACCESS" do
```
jeremyf: Can you go ahead and change throughout the code the constant name:

`Models::AccessRight::RESTRICTED_ACCESS` to `Models::AccessRight::REGISTERED_ACCESS`

laritakr: I changed a few variable names as well. Wasn't quite sure of your intent, but it seemed more consistent to me. Let me know if I should be undoing some of that...

dbrower:
Did every dependency update make its own commit??

jeremyf:
@dbrower TL;DR - Yes.

When I update dependencies in Sipity, I do them one at a time (via the update-dependencies script). This way, if I run into automated test issues, I can use `git bisect` to find which dependency update broke the build.

PR #1137, opened 2017-03-17
Dltp 900 sipity figaro remediation

banurekha:


config/initializers/figaro.rb
```diff
@@ -1,23 +1,47 @@
 Figaro.require_keys(
-  'airbrake_api_key',
```
jeremyf: Why did you change the spacing from 2 to 4 for indentation? The Ruby convention is 2 spaces per indent.

banurekha: Corrected.. Subl Error.. Didn't notice it.

PR #1152, opened 2018-01-17
Fixing batch ingester

jeremyf:
## Fixing batch ingester

ee725aad61abb35a0dfc1287e1787e45319dc197

For Kaminari to work with per page, there is a method call temporal
dependency. First you must call `.page` then you may call `.per`.

The processing query was violating the temporal interface. Below is
the method calls from [the source code][source_code]

```ruby
scope = scope.order(criteria.order) if criteria.order?
scope = scope.page(criteria.page) if criteria.page?
scope = scope.per(criteria.per) if criteria.per?
scope
```

Prior to this change, the `:page` for criteria parameters was nil,
meaning that `scope = scope.page(criteria.page) if criteria.page?` was
skipped. However `:per` was set, and thus
`scope = scope.per(criteria.per) if criteria.per?` was called. This
resulted in a method missing error.

The fix is to amend the batch ingester method to no longer force
pagination. This may mean that we slowly work through a list of objects
15 at a time (or however many we specify).

Addresses DLTP-1230 # Close

[source_code]:https://github.com/ndlib/sipity/blob/47b96299eef5793556cfb45d1a9af27ea1bd992a/app/repositories/sipity/queries/processing_queries.rb#L466-L470

### DIFF
--- a/app/jobs/sipity/jobs/core/bulk_ingest_job.rb
+++ b/app/jobs/sipity/jobs/core/bulk_ingest_job.rb
@@ -70,7 +70,7 @@ module Sipity
 
         def set_search_criteria!
           @search_criteria = search_criteria_builder.call(
-            user: requested_by, processing_state: initial_processing_state_name, work_area: work_area_slug, page: :all
+            user: requested_by, processing_state: initial_processing_state_name, work_area: work_area_slug
           )
         end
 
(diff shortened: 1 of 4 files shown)


dbrower:
Doesn't it make sense to also rewrite that highlighted code block to not call `per` without calling `page` first?

jeremyf:
@dbrower I made changes based on your comment.

PR #1158, opened 2018-02-26
Dltp 967 remediate slow dashboard

jeremyf:
## Adding `rake development:seed_ulra` task

eaae37e27fe26f55a325288964f627a16b65e47f

As I'm interested in having some rudimentary, albeit not very diverse,
data, I wanted to write up methods for seeding data.

## Reducing default pagination from 15 to 8

1b1ca5ead28f3479fe3bd6d493c81d404f9f27b4

There are underlying slowness on the index page due to the flexible data
structure of Sipity and not a direct consideration for the data-
structure necessary for rendering an index view instead of a show page.

This is a quick fix meant to address DLTP-967. If the reduced pagination
does not improve the apparent responsiveness, we'll need to look into
additional measures (e.g. define the data-structure and populate that
with query results)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ SSL=true bin/rails s
 
 Then go to https://localhost:3000
 
+### Development Seeds
+
+`rake bootstrap` - reset the environment's database with basic seed data
+
+In `./lib/tasks/development.rake` you can find Rake tasks for adding data.
+
 ### Work Submission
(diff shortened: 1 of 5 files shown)


jeremyf:
As I'm preparing for deploy on Wednesday, I need to ensure that this change is in place.

PR #1166, opened 2018-06-21
Use batch ingest API in Sipity

laritakr:
DLTP-1380

Add alternate api method for sipity, and set as default.


app/exporters/sipity/exporters/batch_ingest_exporter/api_file_utils.rb
```diff
@@ -0,0 +1,25 @@
...
+        def self.mkdir_p(path)
+          # do nothing
+        end
+
+        def self.mv(_from, destination)
+          response = RestClient.post destination, ""
+        end
+
+        def self.put_file(path, file)
+          response = RestClient.put path, :myfile => File.new(file, 'rb')
```
jeremyf: Prefer `myfile: File.new(file, 'rb')` instead of `:myfile => File.new(file, 'rb')` as the former is the preferred hash syntax in this version of Ruby.

app/exporters/sipity/exporters/batch_ingest_exporter/directory_mover.rb
```diff
@@ -21,14 +20,9 @@ def prepare_destination(path:, file_utility: default_file_utility)
         private_class_method :prepare_destination
 
         def move_files(source:, destination:, file_utility: default_file_utility)
-          file_utility.mv(source, File.join(destination, '/'))
+          file_utility.mv(source, destination)
```
jeremyf: I noticed that the `destination` no longer has a guaranteed appending of `'/'` could you explain the change behind this?

app/exporters/sipity/exporters/batch_ingest_exporter/metadata_writer.rb
```diff
@@ -1,14 +1,38 @@
...
-          FileWriter.call(content: content, path: path)
+          file_writer.call(content: content, path: path)
+        end
+
+        private
+
+        attr_accessor :exporter, :metadata
+
+        def file_writer
+          case exporter.instance_variable_get(:@ingest_method)
```
jeremyf: I'd prefer that you not use `#instance_variable_get`, favor instead to expose the method `#ingest_method` on the exporter.

app/exporters/sipity/exporters/batch_ingest_exporter/webhook_writer.rb
```diff
@@ -24,6 +27,16 @@ def callback_url(work_id:)
           )
         end
         private_class_method :callback_url
+
+        def file_writer(exporter:)
+          case exporter.instance_variable_get(:@ingest_method)
```
jeremyf: I'd prefer that you not use #instance_variable_get, favor instead to expose the method #ingest_method on the exporter.

app/exporters/sipity/exporters/batch_ingest_exporter/directory_mover.rb
laritakr: I wanted a destination that worked for both :api and :files calls. With the api call, the destination can't have the '/' as part of the path, and it works without that for the move, so I removed it rather than adding branching logic.

app/exporters/sipity/exporters/batch_ingest_exporter.rb
```diff
@@ -48,16 +71,30 @@ def with_path_to_data_directory
...
 
+      def job_directory
+        File.join(JOB_PATH.fetch(ingest_method), data_directory_basename)
+      end
+
       def data_directory
-        File.join(DATA_PATH, data_directory_basename)
+        case ingest_method
+        when :files
+          File.join(FILES_PATH.fetch(ingest_method), data_directory_basename)
```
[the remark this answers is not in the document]
laritakr: I changed this. I believe it was a carry-over from some intermediate restructuring I was doing in the code.

PR #1175, opened 2018-07-27
Updating bundle installation protocol to https

jeremyf:
Instead of using git, use https


Gemfile
```diff
@@ -41,6 +48,7 @@ gem 'whenever', require: false
 gem 'rof', github: 'ndlib/rof'
 gem 'rdf-aggregate-repo', '~> 2.0.0'
 gem 'rdf-rdfa', '~> 2.0.1'
+gem 'rdf-isomorphic', '~> 2.0.0' # Required by 2.1
```
jgondron: Did you mean to include this yet?

PR #1183, opened 2018-09-05
Relabeling "Open Access" to "Public"

jeremyf:
## Relabeling "Open Access" to "Public"

0958ec23898e93f6dc632ded5018ee25264f3cdb

DLTP-1542

## Updating language for "ND Only"

1b43d40adb0f3f177cefede697d879f10be4516b

Based on https://github.com/ndlib/sipity/pull/1183 feedback, I'm
updating to replace the language around restricted/registered with
"ND only".

DLTP-1542

### DIFF
--- a/app/views/sipity/controllers/work_submissions/core/access_policy.html.erb
+++ b/app/views/sipity/controllers/work_submissions/core/access_policy.html.erb
@@ -20,10 +20,10 @@
         <tr>
           <th>Type</th>
           <th>Title</th>
-          <th class="single-option">Open Access</th>
+          <th class="single-option">Public</th>
           <th class="single-option">Restricted to Notre Dame</th>
           <th class="single-option">Private</th>
-          <th>Embargo then Open Access</th>
+          <th>Embargo then Public</th>
         </tr>
       </thead>
       <tbody>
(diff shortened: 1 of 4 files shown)


dbrower:
The ticket had a few other things than just changing open access to public.

> Rename the "Open Access" access level in sipity to "Public".
>
> Rename the "Open Access with embargo" to "Public with embargo" (not sure if this appears in sipity, but if it does...)
>
> Rename "Registered" access level to "ND Only". This is the only place in the curatend ecosystem where "registered" appears in a web ui, and Shari has said it is confusing to depositors.

jeremyf:
@dbrower 

I changed "Embargo then Open Access" to "Embargo then Public"; This cleaves close to the conversations we've had that explains the transition of data visibility. I can shift it to "Public with Embargo" but don't know if that further clarifies (or not).

The language for Sipity's "Restricted" is "Restricted to Notre Dame"; I can replace that with "ND Only" but that has been a long-standing language.

I defer to your recommendation as Sipity is very much a Curate "product".

dbrower:
Yes, I prefer the "Embargo then Public". I asked Shari and she prefers ND Only. She says she gets a lot of support questions wondering what Restricted means.

PR #1184, opened 2018-11-16
Send errors via Sentry

laritakr:
Replace Airbrake error handling with Sentry.
Completes DLTP-1502


laritakr:
Key was toggled after initial push.

PR #1187, opened 2018-12-07
Clean blank element in Admin Unit array

laritakr:
Simple-form with multiple entry elements adds an empty element at the beginning of the array. Converting admin unit to multi-element requires hiding the blank element.

See https://api.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html#method-i-select

Related to https://jira.library.nd.edu/browse/DLTP-1594


laritakr:
Note: This same change could probably be used to eliminate `to_array_without_empty_values` and corresponding methods in `degree_form.rb` and other similar situations (where simple_form is used to display a dropdown selectable list from a collection with multiple selections allowed.)

PR #1191, opened 2019-01-15
Modify RestClient put_file

laritakr:
The command was sending the pdf file with the token attached as part of it. The change keeps the token separate.

Fixes DLTP-1611 garbled pdf issue

### DIFF
--- a/app/exporters/sipity/exporters/batch_ingest_exporter/api_file_utils.rb
+++ b/app/exporters/sipity/exporters/batch_ingest_exporter/api_file_utils.rb
@@ -16,20 +16,9 @@ module Sipity
         end
 
         def self.put_file(path, file)
-          # The following tagged the token as part of the file:
-          # RestClient.put path, myfile: File.new(file, 'rb'), 'X-Api-Token' => Figaro.env.curate_batch_api_key!
-          # This appears to transfer the file without adding the token.
-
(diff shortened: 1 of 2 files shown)


app/exporters/sipity/exporters/batch_ingest_exporter/api_file_utils.rb
```diff
@@ -12,11 +12,24 @@ def self.mkdir_p(*)
...
+          # The following tagged the token as part of the file:
+          # RestClient.put path, myfile: File.new(file, 'rb'), 'X-Api-Token' => Figaro.env.curate_batch_api_key!
+          # This appears to transfer the file without adding the token.
+
+          request = RestClient::Request.new(
+          :method => :put,
+          :url => path,
+          'X-Api-Token' => Figaro.env.curate_batch_api_key!,
+          :payload => {
+            :multipart => true,
```
dbrower: Setting multipart to true seems counterintuitive since the batch ingest does not accept multipart encoded requests. But I would believe anything with how restclient operates.

laritakr: I still want to see how batch ingest handles it, but this was one where the request didn't crash AND the token didn't get attached so I committed it to wait until you get back to do more testing.

PR #1202, opened 2019-06-18
Adding command to ease group member management

jeremyf:
When this is merged, I want to update the [dlt-ansible][1] secrets to
use the newly created repository method, as it encapsulates the
logic that keeps the secrets in-sync with the desired state.

Related to DLTP-1633

[1]:https://github.com/ndlib/dlt-ansible/


jeremyf:
@msuhovec Once this is merged, I'll update the Ansible repository to use this new method.

PR #1208, opened 2019-07-17
Addressing shift to ActionController::Parameters

jeremyf:
In moving from Rails 4.x to 5.x, Rails adjusted the parameter object
from a HashWithIndifferentAccess to an ActionController::Parameters
object. The impact is methods that Sipity assumed no longer exist.
This hack address those issues.

In a lament, I look to all of the form objects and how creating a
simple hash to pass to them does not reflect the reality of the
application's actual behavior.

This monkey patch restores the features that I was hoping to continue
to leverage.


dbrower:
Is there a test that would have caught the change in how Rails passes parameters?

Also, why monkey patch a rails class rather than change sipity to use the new interface? I’m thinking of the long-term maintenance costs here.

PR #1209, opened 2019-07-18
Updating schema to address Rails 5.2 realities

jeremyf:
Rails does not adhere to Semantic versioning. Instead they have a
policy in which the minor version can introduce breaking changes. This
usually means prior patch versions will include deprecation notices.

However, in the case of the db/schema.rb file this is not something
that is necessarily vetted. By default the db/schema.rb is only updated
after a migration that has not yet been run is run. When I drop the
local database and run the rake tasks, many things fall apart. This
update address at least the test environment.


msuhovec:
I'm guessing there are no specs for db:migrates

jeremyf:
@msuhovec Nope. I'm working on a 4th commit that addresses the entire teardown and the rebuild for the databases (via `db:drop db:create db:migrate`). I'm also working on shifting test environment from `sqlite3` to `mysql`, to ensure the test database uses the same database driver.

PR #1245, opened 2020-04-14
Favoring prep environment of pre_production

jeremyf:
Replacing references to pre_production with prep. NOTE: this assumes
that the rails applications running on provisioned servers will have
either a `ENV['RAILS_ENV']` of "prep" or "production".

In addition, this commit removes the now defunct staging environment.

Related to DLTP-1815


jeremyf:
When this PR merges, the current (as of 2020-04-14) deploy for the prep environment will break. To fix, it will require work on the ndlib/dlt-ansible scripts for sipity.

jeremyf:
Removing WIP status.

PR #1251, opened 2020-06-15
Accept webhooks with an 'error' status

dbrower:
The webhook endpoint had a validation rejected every webhook except
"success" ones. Thus webhooks for errors in a batch job were being
rejected (with a 422 error), and the code to send a message to Sentry
was never executed. This commit removes the validation.

### DIFF
--- a/app/forms/sipity/forms/work_submissions/core/ingest_completed_form.rb
+++ b/app/forms/sipity/forms/work_submissions/core/ingest_completed_form.rb
@@ -24,11 +24,14 @@ module Sipity
           end
 
           include ActiveModel::Validations
-          validates :job_state, inclusion: { in: [JOB_STATE_SUCCESS] }
 
           def submit
+            if job_state == JOB_STATE_SUCCESS
+              processing_action_form.submit { create_a_redirect }
+              return true
+            end
             register_error if job_state == JOB_STATE_ERROR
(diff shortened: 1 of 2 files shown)


jeremyf:
Reviewing the Sentry logs, we did receive Sipity notifications (and may need to look into why those were not as visible).

I believe the better approach might be to move the result into an error state. This would give visibility to Shari and would allow us to remediate. I'm going to close this and propose a discussion on how to proceed.

dbrower:
The problem still remains that the endpoint responds with a 422 status with the routine webhooks saying a job is starting or that there is a processing error. Can it at least reply with a 200? Then it would be meaningful when the batch system detects a non-200 response. see https://github.com/ndlib/curatend-batch/pull/139

dbrower:
I tried to update the PR but I'm having a hard time figuring out how to do it. There is a lot of indirection!

jeremyf:
@dbrower I want to hold off on this. Until we identify how Sipity handles an error status from the WEBHOOK, we should not be trying to make changes.

If we say "Yes Sipity will advance the state based on an error" then we can adjust accordingly. If we say "No, Sipity will not advance the state based on an error" then we have a different approach.

dbrower:
Waiting to figure out the error handling is fine.

But can the "processing" messages be addressed? I think it is clear those shouldn't change the state.

PR #1253, opened 2020-06-17
Reworking IngestCompleted to handle error state

jeremyf:
Prior to this commit, there was an assumption that the IngestCompleted
form was only used to advance the state.  However, in discussion with
Don, the purpose of the WEBHOOK for the batch ingester was meant to be a
communication tool (e.g., post status to WEBHOOK when we start
processing, when there's an error, and when there's success).

Under the above batch ingester assumption, the client should not send a
422 because ERROR was invalid.  Instead, it should send a 200
status (e.g. yup, I got your message).

Future work might include reworking the form to handle routing to
different states based on the status of the WEBHOOK.  For now, this gets
things unstuck.


dbrower:
Does this also need to check for the “processing” state as being valid?

jeremyf:
Perhaps. Is that also throwing errors?

jeremyf:
@dbrower I added the `processing` state.

PR #1264, opened 2021-02-26
Request a DOI for each ETD submission

laritakr:
1. Add DOI attribute to ETD work types with default value "To be assigned"
2. Incorporate request to mint DOI through batch ingest.

CURATE-248

Also completes CURATE-284


msuhovec (commented):
OK, I am missing something. I see the TaskWriter module, and how it will build the JOB file with the designated taskname in the todo list. Where is this called, and what is that task name for doi minting?

msuhovec (changes requested):
Could you change the task name from doi-request to start-doi-request?  All of the initial workflow tasks (so far) begin with 'start'. That's what the initial task for doi-minting is named at present.

laritakr:
> OK, I am missing something. I see the TaskWriter module, and how it will build the JOB file with the designated taskname in the todo list. Where is this called, and what is that task name for doi minting?

This is just a WIP PR because I'm going over some questions tomorrow.... more changes will be coming.

msuhovec (approved):

PR #1287, opened 2021-08-11
Adding Author Name and Submission Date to Dashboard

jeremyf:
## Interstial change to add two attributes to dashboard

3103308adf88e67f2848d12dbc112074833326d7

I'm not certain that this will work, but I want to capture the logic
that I've done to generate the joins.

The reason that this is not quite right is that I need to add the change
into given `criteria` AND `scope_proxied_objects_for_the_user_and_proxy_for_type` method.

The `criteria` object defines what are the sort columns and query
strings; so I need to expose the field sort for that.  The
`scope_proxied_objects_for_the_user_and_proxy_for_type` creates the
initial AREL scope from which we then build out additional queries.

So, while this "kind of works" and "kind of" addresses the
situation (e.g., the product owner wants to see the author and the
submission date in the dashboard and the dashboard is too slow), it's
not yet done.

## Switching from WHERE to ON in JOIN

08c4209c6a1c67a20eeed5e18d78218396f14bfb

Prior to this, the WHERE required the existence of the joined row.  In
moving it to ON, the query now allows for a NULL author_name or
submission_date

## Moving methods onto search criteria

641b1d5618e59bb5674dc20f62ba742a0e91b854

This is a bit of an odd duck, the criteria parameters are there to help
consolidate the logic necessary to run a search/filter/sort query for
the dashboard.  So there are two inter-related things:

1) The Active Record scope object
2) The criteria parameters

And one updates the other; regardless of how I do this.  I could extract
the logic to happen within the criteria (and am tempted given that I'm
now blending SQL logic into that criteria).

Ultimately, the goal is to allow for the querying of the dashboard to:

1) include a sort on an additional attribute (e.g. submission_date)
2) reduce the number of ActiveRecord queries necessary for building out the
   dashboard

This change iterates in that direction.

What remains?

- [ ] adding the sort criteri
- [ ] updating the templates to render those fields
- [ ] ripping out some of the Curly view object’s antics
- [ ] perhaps removing an `includes` directive on the dashboard's query

## Adding protection against SQL injection

d1704c7363b899ffce65be03a796a0ee5fdbfbdc

Prior to this commit, the SearchCriteriaForWorksParameter class was to
optimistic in the additional_attributes it received.  This was "okay" in
that these attributes were passed from within the application (and not
from user input).

However, I wanted to tighten that up further which is why I added the
changes to fetch the attributes from a map.

And even with all of this, [brakeman][1] identified this as a weak
chance for SQL injection.  So I also added a config/brakeman.yml file to
skip over this file. (alas I can't skip over a singular violation)

[1]:https://rubygems.org/gems/brakeman

## Adding additional attributes for /area/etds

2dbec3660a76390a0bf830f382f2dd342102a47e

While technically not necessary, I wanted to be specific in declaring
what attributes we're including.

I also took the time to refactor the public/private declarations.  It's
a bit more idiomatic.

## Adding author_name and submission_date to etds show

9662ba7399fafe658c47b29ead70d6e4fb00c6df

This builds on the work in the
`Sipity::Parameters::SearchCriteriaForWorksParameter` in which we add
the `author_name` and `submission_date` to the returned ActiveRecord
objects.

This reduces the number of queries that we make.

## Returning the modified scope instead

332e7f234ac15e7f5aa8e01d265fe7fb502108b3

Prior to this change, the code modified the select statement but did not
return that modification.  With this change the select is applied to the
scope and that modified scope is returned.  (Thus the upstream queries
include the virtual attributes).

## Fixing parameter name for dashboard

a8b8c39a3c1f9f9484fac1374e9f0c63c7a834f2

Prior to this, the `submission_date` was not a defined additional
attribute.  So I changed it to `etd_submission_date` (which is defined).

I've tested these changes in the UI to see the sorting mechanisms.

## Refactoring to repurpose a common method

0f66475a1b3a275f2224b1c5ced8a40245103829

Closes https://jira.library.nd.edu/browse/CURATE-288

## Removing author sort

f4291400a32d1f456a7d938a02890c3ec3390649

This really isn't all that helpful, so we'll leave it out.

## Removing author name from the order by options

f39d78a5829f1d41d335af4b5fe4a206a718b0c6

This is not all that useful.  And clutters the interface.

## Improving performance of dashboard response

8ad8ab86d4a738be8eb2c9c2a284c1d870679791

Prior to this commit, for each work on the dashboard, we'd query the
additional parameters for the program names.  With this change, we're
defaulting the dashboard query to add the program_name(s) to the initial
query result set.  This removes one of the conditions of N+1 queries.

One small concession: the SQL concatentation is with a ", " instead of
using Rails's `to_sentence`.  The result is that the array `["A", "B",
"C"]` goes from "A, B, and C" to "A, B, C".

## Having ShowPresenter specify additional_attributes

9c353f8ff67a0e1e420bc39cc6f5a78f4103054d

Prior to this commit, I made the assumption that the ShowPresenter would
defer to the `SearchCriteriaForWorksParameter` for the additional
attributes.

However, as part of the PR, LaRita recommended that each context should
specify it's additional attributes.  Which is spot on.

This change now, in essence, makes the prior changes an additive change
that by default doesn't impact other implementations.  And instead
updates the dashboard rendering behavior to be the only place where the
additional attributes of "author_name", "etd_submission_date", and
"program_name" might be applied.

### DIFF
--- a/app/parameters/sipity/parameters/search_criteria_for_works_parameter.rb
+++ b/app/parameters/sipity/parameters/search_criteria_for_works_parameter.rb
@@ -17,6 +17,22 @@ module Sipity
       self.default_q = nil
       self.default_order = 'title'.freeze
       self.default_additional_attributes = ["author_name", "submission_date"].freeze
+
+      # Note the parity between this and the additional attributes.
+      # I'm including the following map to remove a possible SQL
+      # injection spot.
+      #
+      # Each key has a value that is a hash with two keys:
+      # 1. `key`: - the `sipity_additional_attributes.key` field's value
+      # 2. `join_as_table_name`: - for the purposes of the join, the
+      #    table name we'll join as.
+      #
+      # We need the `join_as_table_name` so that we can join multiple
+      # times to the same `sipity_additional_attributes` table.
+      ADDITIONAL_ATTRIBUTE_MAP = {
+        "author_name" => { key: 'author_name', join_as_table_name: 'author_names' },
+        "submission_date" => { key: 'submission_date', join_as_table_name: 'submission_dates' },
+      }
       ORDER_BY_OPTIONS = ['title', 'title DESC', 'created_at', 'created_at DESC', 'updated_at', 'updated_at DESC'].freeze
 
       def self.order_options_for_select
@@ -68,11 +84,12 @@ module Sipity
         select_fields = ["#{work_table_name}.*"]
 
         additional_attributes.each do |attribute|
-          table_name = attribute.to_s.pluralize
+          table_name = attribute.fetch(:join_as_table_name)
+          key = attribute.fetch(:key)
           scope = scope.joins(
-            %(LEFT OUTER JOIN #{attr_table_name} AS #{table_name} ON #{table_name}.work_id = #{work_table_name}.id AND #{table_name}.key = "#{attribute}")
+            %(LEFT OUTER JOIN #{attr_table_name} AS #{table_name} ON #{table_name}.work_id = #{work_table_name}.id AND #{table_name}.key = "#{key}")
           )
-          select_fields << "#{table_name}.value AS #{attribute}"
+          select_fields << "#{table_name}.value AS #{key}"
         end
 
         scope.select(select_fields.join(", "))
@@ -84,7 +101,7 @@ module Sipity
       # Doing our due diligence to santize parameters
       def additional_attributes=(input)
         @additional_attributes = Array(input).map do |attribute_name|
-          Models::AdditionalAttribute.sanitize_sql_for_conditions(attribute_name.to_s)
+          ADDITIONAL_ATTRIBUTE_MAP.fetch(attribute_name)
         end
       end
 
--- a/app/parameters/sipity/parameters/search_criteria_for_works_parameter.rb
+++ b/app/parameters/sipity/parameters/search_criteria_for_works_parameter.rb
@@ -92,8 +92,8 @@ module Sipity
           select_fields << "#{table_name}.value AS #{key}"
         end
 
+        # Note this must return the modified scope.
         scope.select(select_fields.join(", "))
-        scope
       end
 
       attr_writer :user, :processing_state, :proxy_for_type, :work_area, :submission_window, :per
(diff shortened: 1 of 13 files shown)


app/presenters/sipity/controllers/work_areas/core/show_presenter.rb
```diff
@@ -7,23 +7,32 @@ module WorkAreas
...
-                repository: repository, work_area: work_area, q: work_area.q, submission_window: work_area.submission_window
+                user: current_user,
+                processing_state: work_area.processing_state,
+                page: work_area.page,
+                order: work_area.order,
+                repository: repository,
+                work_area: work_area,
+                q: work_area.q,
+                submission_window: work_area.submission_window,
+                additional_attributes: ["author_name", "etd_submission_date"]
```
laritakr, line 35: Should the attributes be a hardcode part of core if they include "etd_submission_date"? Should it go into etd/show_presenter?

app/parameters/sipity/parameters/search_criteria_for_works_parameter.rb
```diff
@@ -16,7 +16,31 @@ class SearchCriteriaForWorksParameter
       self.default_submission_window = nil
       self.default_q = nil
       self.default_order = 'title'.freeze
-      ORDER_BY_OPTIONS = ['title', 'title DESC', 'created_at', 'created_at DESC', 'updated_at', 'updated_at DESC'].freeze
+      self.default_additional_attributes = ["author_name", "etd_submission_date"].freeze
```
laritakr, line 19: I don't think we want to sort by author name

laritakr (changes requested):

app/parameters/sipity/parameters/search_criteria_for_works_parameter.rb
jeremyf: I went ahead and removed the author_name from the sort.

app/presenters/sipity/controllers/work_areas/core/show_presenter.rb
jeremyf: Good catch.  I'll rework to parameterize accordingly.

jeremyf: I have parameterized as part of my [last commit](https://github.com/ndlib/sipity/commit/bc78b4b9e21a8501bd01c018e0c11a7eb22f5862/)

laritakr (approved):
Looks good so far... were you planning to test this out before changing the per page limit, and then doing that separately?

jeremyf:
I've tested with the data that I have.  The next step is to test responsiveness on PREP.  Finally, change the pagination size if it appears to resolve things.

PR #1292, opened 2021-08-16
Adding default omit terminal actions

jeremyf:
## Interstitial commit to query multiple states

2400ed74057e29d7067dc88e1f783618965e6882

With this commit, you can now select and filter on multiple states on
the dashboard.

## Switching work area interface

4e8ce0b7caea1a51ca6c70b3b002748873d375fe

This aligns with the emerging changes for the
`SearchCriteriaForWorksParameter` interface (e.g., moving from single
value to multiple value).

## Updating work_area to be an "first set" attribute

3a0b76421a30487679466e15b4d7f1c16a6b45f4

Prior to this commit, the order of assignment was somewhat arbitrary.
And as I want to amend the default_processing_states to start with
non-terminal states, I need to ensure that work_area is set before the
others (so that I can query the work area's non-terminal states).

## Renaming search/filter parameter to plural

a16ec11bdf9429de594fde662f2115bcdf0c1262

Prior to this change, the parameter name presumed a singular value.
With this change, the parameter name implies plurality.

As part of the refactor, I've removed the singular name (which could
remain if we alias plural to singular, but instead I chose to remove it
for clarity).

## Parameterizing list for states to allow non-terminal

981ad969d4d7dd68ee661b3ef8796e659fc849b1

This commit is a refactoring of a query, to allow for more immediate and
simplified AREL query chaining.

The goal of this change is to introduce a means for generating a list of
non-terminal states for the given work area.  Non-terminal states are
those states that have actions that can transition them out of the given
state.  A terminal state is an a state in which there are no actions
that will transition the object out of it's current state.

## Removing "all states" from select list

f3d97eda4b0023461da4c1a17224b72d33f671c6

In local testing, the Rails logic around the input prompt and
multi-select can create some inconsistent behavior.

For example, if I selected only "All States", then ran the search, the
returned page would not indicate that "All States" was selected.  In
other words, the select box would have nothing selected.  I found that
this created an odd visual state.

Thinking about this it makes sense; what does it mean to have a
multi-select and then someone selects "All States" and "New"?  Which do
we honor in the filter?

My response is to remove the "All States" prompt/placeholder thus
showing specifically which states you've selected.  I suppose if you
select NONE of them, the inner logic will render all of the states.

Addresses - https://jira.library.nd.edu/browse/CURATE-289

### DIFF
--- a/app/repositories/sipity/queries/processing_queries.rb
+++ b/app/repositories/sipity/queries/processing_queries.rb
@@ -520,11 +520,11 @@ module Sipity
           returning = entities[:proxy_for_type].eq(proxy_for_type).and(
             responsibility[:actor_id].in(user_actor_contraints)
           )
-          if criteria.processing_state?
+          if criteria.processing_states?
             returning = returning.and(
               entities[:strategy_state_id].in(
                 strategy_states.project(strategy_states[:id]).where(
-                  strategy_states[:name].eq(criteria.processing_state)
+                  strategy_states[:name].in(criteria.processing_states)
                 )
               )
             )
--- a/app/repositories/sipity/queries/processing_queries.rb
+++ b/app/repositories/sipity/queries/processing_queries.rb
@@ -35,9 +35,12 @@ module Sipity
       #
       # @param work_area [Object] that can be converted into a Sipity::Models::WorkArea
       # @param usage_type [Object] the polymorphic type for database storage
+      # @param include_terminal [Boolean] when true, include
+      #        processing states that are terminal (e.g., there are no
+      #        actions that transition out of the state).
       #
       # @return [Array<String>] name of actions available
-      def processing_state_names_for_select_within_work_area(work_area:, usage_type: Sipity::Models::WorkType)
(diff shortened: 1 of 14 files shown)


app/repositories/sipity/queries/processing_queries.rb
```diff
@@ -46,19 +49,29 @@ def processing_state_names_for_select_within_work_area(work_area:, usage_type: S
...
-            strategy_usages[:usage_type].eq(usage_type)
-          )
-        ).join(submission_window_work_types).on(
-          submission_window_work_types[:work_type_id].eq(strategy_usages[:usage_id])
-        ).join(submission_windows).on(
-          submission_windows[:id].eq(submission_window_work_types[:submission_window_id]).and(
-            submission_windows[:work_area_id].eq(work_area.id)
-          )
-        )
+        scope = Models::Processing::StrategyState.distinct.
```
laritakr, line 52: My brain is spinning a bit from this query, but it seems to work.....

laritakr (approved):

PR #1293, opened 2021-08-17
Removing code climate reporting

jeremyf:
## Removing code climate reporting

059afc6b151b7a3b7bb19cb010f750a059b16423

Prior to this commit, the Github CI checks included a step that pushed
the code coverage report to CodeClimate.  After this commit, that step
is gone.

NOTE: the build will still break if we do not meet the coverage
requirements.


jeremyf:
Note: I pushed up a previous change with a speculative message.  That change did what I'd hoped (and wrote in the updated commit message).  However that test needed to be re-run, so we're again waiting.

laritakr (approved):

PR #1299, opened 2021-09-29
Enlarge page size for exporter queries

laritakr:
Exporters only handle one page of items. Rather than expend a large amount of time to handle pagination in the exporters, it is much easier to just override the default page size for exporter queries.

Resolves CURATE-342


msuhovec (approved):
we will hope to God that wer never live to see the day with more than a thousand ETD submissions!

PR #1304, opened 2021-10-19
Minor update to get mysql2 running in dev - minor change to Gemfile

rfox2:
Pointing some gems from private repo to the ndlib repo, and upgraded mysql2 gem


laritakr:
Change breaks build, so closing PR. Can be reopened in the future if someone wants to make needed revisions.

PR #1307, opened 2021-12-09
task: Remove puma, update bundle and ci spec

msuhovec:


Gemfile.lock
```diff
@@ -8,7 +8,7 @@ GIT
 
 GIT
   remote: https://github.com/jeremyf/data-migrator.git
```
laritakr, line 10: You appear to not be using the current Sipity codebase for this work, because we had previously moved to using this gem from ndlib rather than jeremyf

Gemfile.lock
```diff
@@ -485,12 +478,11 @@ GEM
       binding_of_caller (~> 1.0)
       pry (~> 0.13)
     public_suffix (4.0.6)
-    puma (3.12.6)
```
laritakr, line 488: I made this change in the curate-285-replace-file branch

Gemfile.lock
msuhovec: I closed this pull request and deleted the branch- when we've merged your replacefile stuff, I will start twith that as a base.

PR #1310, opened 2022-04-01
Use Entity Registry to track file replacements

laritakr:
DAS-3027 (Related to https://github.com/ndlib/sipity/pull/1306) 

After a file is replaced and re-ingested, the "Replace File" to-do item was marked as complete, and the "Reingest File" action button is active. 

Since file replacement can be repeated multiple times, this problem is solved by removing the Entity Action Registration for "replace file" once the ingest is complete as part of the callback from batch ingest.


msuhovec (approved):

laritakr:
Note that this also changed the method used to determine which file(s) to send to batch ingest, to use registrations rather than activity logs so they are both in synch.

PR #1312, opened 2022-05-12
Update mysql2 gem

rfox2:
These changes are primarily the result of getting the app to build on a new system but they are minor changes. Please review the update to the schema.rb file - that change happened when I ran the database creation commands.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -32,7 +32,7 @@ local_user: &local_user
   <<: *mysql_connection
   username: root
 <% if ! ENV['SKIP_MYSQL_PASSWORD_FOR_LOCAL_DEVELOPMENT'] %>
-  password: root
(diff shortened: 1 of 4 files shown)


config/database.yml
```diff
@@ -32,7 +32,7 @@ local_user: &local_user
   <<: *mysql_connection
   username: root
 <% if ! ENV['SKIP_MYSQL_PASSWORD_FOR_LOCAL_DEVELOPMENT'] %>
-  password: root
+  password:
```
laritakr, line 35: Password is required for CI. To run locally, you can set the environment variable.

laritakr (changes requested):

laritakr:
@rfox2 Changes are required here.

PR #1314, opened 2022-07-29
lam-3091 use locabulary 0.82

msuhovec:


abbales:
Why are the ref elements removed? Will this improve the process for future changes?

msuhovec:
Honestly, there are several different ways to do it.

If there is no github reference and no gem version on the line, the latest
version of the gem is fetched - you can see in the Gemfile.lock that 0.8.2
was used.

One can also specify the version, or branch, or github reference that the
gem was built against.

If you look at the other ndlib gems included in the Gemfile, we are not
consistent- we use all of these methods.

I usually either use the gem version or just let it fetch the latest. I
could not think of a good reason why I would not want the latest, so I went
with that method

abbales:
Using the latest version seems like a good idea to me. I can't think of a reason we would ever want something other than the latest

abbales (approved):

msuhovec:
Agreed!

